PR #22, opened 2016-07-26
Add RotateFile as Log Handler

maurorodrigues:
With RotateFileHandler the log file will be limited to 2MB and 1 backup kept.


maurorodrigues:
Change tested and logs limited as expected! LGTM

PR #29, opened 2016-08-23
Fix scripts failing to produce results with newly created user

ghost:
Those patches fix the first execution when a single package is selected and make sure the mock executable does not require root privilege.


maurorodrigues:

lib/mockbuilder.py
```diff
@@ -26,6 +26,7 @@
 CONF = config.get_config().CONF
 LOG = logging.getLogger(__name__)
 MOCK_CHROOT_BUILD_DIR = "/builddir/build/SOURCES"
+MOCK_BIN = "/usr/bin/mock"
```
sathnaga, line 30: @olavphibm How about finding installed location of mock if not failout?

maurorodrigues:
good to go! Thanks for the patch!

PR #85, opened 2016-09-30
Fix name shadowing in get_reference method

farosas:
The 'exception' variable shadows the name of the module in the outer
scope, raising an AttributeError.


maurorodrigues:

ghost (approved):
Thanks!

PR #90, opened 2016-10-10
Alter precedence of configuration options sources

### LINKED ISSUE #83: Respect log_file value set in config.yaml
The value of ``log_file`` that is set in ``config.yaml`` is always overwritten by a default value in ``lib/config.py``, so ``log_file`` value in ``config.yaml`` is not respected.

A desirable precedence should be (least precedence listed first):

- ``config.yaml``
- default value in ``lib/config.py`` if ``log_file`` does not exist in ``config.yaml``
- ``--log-file`` switch

farosas:
Issue #83 points out an inconsistent behavior in the argument-parsing code in which values in the configuration file have no effect if a default is set in argparse (via the `default` keyword in `add_argument`) .

Fixes #83

### DIFF
--- a/lib/config.py
+++ b/lib/config.py
@@ -71,14 +71,14 @@ def discover_software():
 
 
 class ConfigParser(object):
+    """
+    Parses configuration options sources.
 
+    Precedence is:
+    cmdline > config file > argparse defaults
+    """
     def __init__(self):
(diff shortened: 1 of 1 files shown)


maurorodrigues:

ghost (approved):
Looks good

PR #91, opened 2016-10-11
Avoid discovering packages before versions repo checkout

### LINKED ISSUE #54: host-os-build.py crashes and throws exception when changing build version
**How to reproduce**
Given for instance current versions master, which contains versions for sos, and a version tag which doesn't (none in special, just removed sos, committed tagged and used it as build version).

```
[user@jaspion1 builds]$ python host-os-build.py --package sos --build-version another
Logs available at /var/log/host-os/builds.log
Traceback (most recent call last):
  File "host-os-build.py", line 100, in <module>
    sys.exit(main(sys.argv[1:]))
  File "host-os-build.py", line 92, in main
    return build_manager()
  File "/home/user/dev/builds/lib/manager.py", line 43, in __call__
    self.packages = self._prepare_packages()
  File "/home/user/dev/builds/lib/manager.py", line 76, in _prepare_packages
    self.packages_list)]
  File "/home/user/dev/builds/lib/package.py", line 51, in __init__
    self.load_package(package, distro)
  File "/home/user/dev/builds/lib/package.py", line 68, in load_package
    with open(self.package_file, 'r') as package_file:
IOError: [Errno 2] No such file or directory: '/home/user/dev/builds/components/sos/sos.yaml'
```

This happens cause when host-os-build.py start it had sos directory, but then it checks out into the selected build version which removes it and things blow up.

farosas:
Fixes #54

### DIFF
--- a/host-os-build.py
+++ b/host-os-build.py
@@ -93,7 +93,6 @@ def main(args):
             conf.get('default').get('build_versions_repository_url'),
             config.COMPONENTS_DIRECTORY, version)
 
-        # rediscovery software if it was not set
         conf['default']['packages'] = conf['default']['packages'] if (
             conf.get('default').get('packages')) else (
                 config.discover_software())
--- a/lib/config.py
+++ b/lib/config.py
@@ -91,7 +91,6 @@ class ConfigParser(object):
         return conf
 
     def _parse_arguments(self):
-        supported_software = discover_software()
         parser = argparse.ArgumentParser()
         parser.add_argument('--config-file', '-c',
                             help='Path of the configuration file for build '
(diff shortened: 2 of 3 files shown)


maurorodrigues:

ghost (commented):
Although I approve the patch, I don't think it fixes issue #54. You are still raising an exception.
The problem is that the packages to build are discovered in config.py, and at that moment there might be a previously checked out branch in components/. Later on, in host-os-build.py, the correct branch is checked out, and, if a previous branch/tag is used, it might not contain one of the packages. It will then break when trying to load that package.

farosas:
@olavphibm I think we're discussing different issues here. In #54, `--package sos` is used, so there is no influence from `config.discover_software`.

EDIT: just discussed with @olavphibm about this. I'll add some patches to fully address the issue.

ghost (approved):
It looks like this solves the issue.

Issue #82, opened 2016-09-29
Add switch to specify repositories_path value


mopsfelder:
The value of ``repositories_path`` is only set in ``config.yaml``.

This request is to add a new command line switch in ``host-os-build.py``, e.g. ``--repos-path``, to allow setting its value.

snowcrshd:
I can work on this.

Would the command line argument have precedence over the configuration file?

olavphibm:
@snowcrshd Yes, like mentioned here https://github.com/open-power-host-os/builds/pull/90/files

snowcrshd:
@olavphibm  cool

fabianorosas:
@snowcrshd's patch fixes this issue.

PR #92, opened 2016-10-14
Refactor ConfigParser,  add unit tests for it and --repositories-path argument

ghost:
This should solve issue #82

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -150,3 +150,23 @@ Validating
 
 There is a whole repository dedicated to testing available at
 https://github.com/open-power-host-os/tests
+
+Running unit tests
+------------------
+In order to run the build scripts unit tests, you will need to install our
+development dependencies.
+
+You can do this by issuing the command below
+
+::
+
+$ sudo pip install -r requirements-dev.txt
+
+
+From the root of the `builds` project directory, use the commands below to run
+the unit tests:
+
+::
+
+$ export PYTHONPATH=$(pwd):$PYTHONPATH
+$ nosetests tests/unit
--- a/lib/config.py
+++ b/lib/config.py
@@ -110,6 +110,9 @@ class ConfigParser(object):
         self.parser.add_argument('--result-dir', '-r',
                                  help='Directory to save the RPMs.',
                                  default='./result')
+        self.parser.add_argument('--repositories-path', '-R',
+                                 help='Directory where to clone code repositories',
+                                 default='/var/lib/host-os/repositories')
         self.parser.add_argument('--keep-builddir',
                                  help='Keep build directory and its logs and '
                                  'artifacts.', action='store_true')
--- a/lib/config.py
+++ b/lib/config.py
@@ -16,6 +16,7 @@
 import argparse
 import logging
 import os
+import sys
 
 import yaml
 
@@ -29,6 +30,7 @@ def get_config():
     global config_parser
     if not config_parser:
         config_parser = ConfigParser()
+        config_parser.parse()
     return config_parser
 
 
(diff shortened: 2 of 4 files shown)


maurorodrigues:

lib/config.py
```diff
@@ -77,62 +79,88 @@ class ConfigParser(object):
...
+        """
+        Parses the arguments provided in the argument list and returns
+            the result object.
+        """
+        result = self.parser.parse_args(args)
+        return vars(result)
+
+    def parse_config_file(self, config_file_path):
+        """
+	Parse the configuration file and returns a dictionary containing the
```
ghost: Indent with one more whitespace.
"return" instead of "returns"

lib/config.py
```diff
@@ -77,62 +79,88 @@ class ConfigParser(object):
...
-                                 'repository')
+        config_file = self.parser.parse_known_args()[0].config_file
 
-        config = self._parse_config(config_file)
-        parser.set_defaults(**config['default'])
+        config = self.parse_config_file(config_file)
+        self.parser.set_defaults(**config['default'])
 
-        args = vars(parser.parse_args())
+	args = self.parse_arguments_list(sys.argv[1:])
```
ghost: Indent one more whitespace.

lib/config.py
```diff
@@ -77,62 +79,88 @@ class ConfigParser(object):
...
+                                 nargs='*')
+        self.parser.add_argument('--log-file', '-l',
+                                 help='Log file',
+                                 default='/var/log/host-os/builds.log')
+        self.parser.add_argument('--verbose', '-v',
+                                 help='Set the scripts to be verbose',
+                                 action='store_true')
+        self.parser.add_argument('--result-dir', '-r',
+                                 help='Directory to save the RPMs.',
+                                 default=os.path.join(self._cwd, 'result'))
```
ghost: This should be treated the same as '--repositories-path', '--log-file' and '--config-file'.
To make them testable we could either keep as is, which could maybe cause problems when passing to other programs (mock?) or expand the full path, which would require tweaking the tests.
This would allow removing the parameter cwd from the constructor, which I think is unnecessary.

ghost (changes requested):
Personally, I'd split the first commit into 3, one for each item you mention. I think that makes review easier by simplifying each diff.

lib/config.py
ghost: I did not understand your point.
What should be done to the options you mentioned?

I also did not how your solution would work, regarding other programs.
Could you please provide an example?

lib/config.py
ghost: That's actually a tab :smile: 
Thanks for pointing that out.

README.rst
```diff
@@ -150,3 +150,23 @@ Validating
...
+development dependencies.
+
+You can do this by issuing the command bellow:
+
+::
+
+$ sudo pip install -r requirements-dev.txt
+
+
+From the root of the `builds` project directory, use the commands bellow to run
```
farosas: below

lib/config.py
farosas: @snowcrshd I suggest you make a pass through pycodestyle just to make sure no more issues like this go unnoticed.

lib/config.py
ghost: The points I am trying to make are the following:
- I dislike the parameter cwd in the constructor. The user should either pass parameters relative to the current working dir or absolute dirs. Furthermore, passing an arbitrary cwd would require also making the paths got from config file relative to that cwd and I believe this disrupts the concept of "current" working dir.
- I believe the point of having os.path.join(self._cwd, 'result')) is to create an absolute path. Since this parameter is passed to mock, and the cwd could change along the process, using a relative path here could probably break things. os.path.abspath could be used instead.
- If we are converting the result_dir to an absulte path, why not convert the other paths too?
- If you agree with the points above, the tests will need to be reworked to check absolute paths.

lib/config.py
farosas: Heretic!

lib/config.py
ghost: TBH,  I don't like the idea of our code implicitly converting paths to absolute paths in these cases.

I think it is better to allow the user to give us an absolute path if this is what they wish.

> passing an arbitrary cwd would require also making the paths got from config file relative to that cwd and I believe this disrupts the concept of "current" working dir.

Why so? If we read the paths from the configuration file, I don't see why we should apply any changes to them. The user should provide the correct paths in this case.

> Since this parameter is passed to mock, and the cwd could change along the process, using a relative path here could probably break things. os.path.abspath could be used instead.

How could the be a problem? If the CWD changes, the attribute in ConfigParser will still hold
the value we are interested in.

Maybe converting that CWD argument to an absolute path would be best.

> If you agree with the points above, the tests will need to be reworked to check absolute paths.

lib/config.py
ghost: lol, had to spin up a new VM and forgot to configure my editor for that :P

ghost:
@olavphibm @fabianorosas I split the first commit into three, this should make reviewing the patch easier.

lib/config.py
ghost: > TBH, I don't like the idea of our code implicitly converting paths to absolute paths in these cases.

What's the problem? The resulting path will be pointing to the same place. Anyway, I've checked and using a relative path for result_dir does not cause problems, so we could just leave the relative path.
The problem I see with the current implementation is the different behavior from command line and config file.
If current_working_dir="/foo" is passed to the constructor and '--result-dir=bar' is passed to the command line, result_dir will be "/foo/bar in the resulting dict.
However, if result_dir is instead passed as 'bar' in the config file, the constructor argument will not be prepended, and result_dir will be just "bar in the resulting dict.
I think this "os.path.join(self._cwd," should just be removed.

ghost: Ah, now that's clear.
That example helped me see the problem.

Well, I'll remove that os.path.join then and remove the _cwd attribute

ghost: I think this is all that's needed, sorry for overcomplicating things :P

ghost (approved):

PR #95, opened 2016-10-19
Fix exception handling in package.py

farosas:
The issue was introduced by: "8e32920 Catch exception from missing
package files in target branch"


maurorodrigues:

maurorodrigues (approved):

PR #96, opened 2016-10-19
Allow automatic updating of spec files version and release fields

farosas:


maurorodrigues:
Do not trigger tests against this one, the usual test workflow doesn't include the upgrade_version.py. Let's test it locally and give or not our approvals

farosas:
Depends on https://github.com/open-power-host-os/versions/pull/43

farosas:
Just a heads-up to anyone testing this, the script still needs to have the components directory already present. I'll alter it to clone the versions repo if it is not found but it will require some refactoring in host-os-build.py so it might take a bit of time.

upgrade_versions.py
```diff
@@ -15,75 +15,212 @@
...
+        self._bump_release(pkg, changelog)
+
+    def _update_version(self):
+        LOG.info("%s: Updating version to: %s" % (self.pkg,
+                                                  self._repo_version))
+
+        with open(self.pkg.specfile, 'r+') as f:
+            content = f.read()
+
+            version = re.search(r'Version:\s*(\S+)', content).groups(1)[0]
```
farosas: This will always work, but it is wrong. It should be `.group(1)` instead of `.groups(1)[0]`

upgrade_versions.py
```diff
@@ -82,33 +85,142 @@ def rpm_cmp_versions(v1, v2):
...
+        LOG.info("%s: Current version: %s" % (self.pkg, self._spec_version))
+        self._spec_release = rpm_query_spec_file(
+            'release', self.pkg.specfile).split('.')[0]
+
+    def _read_version_from_repo(self, repo_path):
+
+        version_regexes = [
+            self.pkg.version_file_regex,
+            ('VERSION', r'.*'),
+            ('%s.spec' % self.pkg, r'Version:(.*)'),
```
ghost: There might be whitespaces after ":":  `r'Version:\s*(.*)'`

upgrade_versions.py
```diff
@@ -82,33 +85,142 @@ def rpm_cmp_versions(v1, v2):
...
+                    match.group().strip())
+                LOG.info("%s: Repository version: %s" % (self.pkg,
+                                                         self._repo_version))
+                break
+            except:
+                pass
+        else:
+            LOG.error("%s: Could not find version in the source repo." %
+                      self.pkg)
+            msg = "Could not find %s's version in the source repo." % self.pkg
```
ghost: You could pass this msg to LOG.error

upgrade_versions.py
```diff
@@ -82,33 +85,142 @@ def rpm_cmp_versions(v1, v2):
...
+        self._spec_version = rpm_query_spec_file('version', self.pkg.specfile)
+        LOG.info("%s: Current version: %s" % (self.pkg, self._spec_version))
+        self._spec_release = rpm_query_spec_file(
+            'release', self.pkg.specfile).split('.')[0]
+
+    def _read_version_from_repo(self, repo_path):
+
+        version_regexes = [
+            self.pkg.version_file_regex,
+            ('VERSION', r'.*'),
```
ghost: Enclose in parenthesis so it matches a group

upgrade_versions.py
```diff
@@ -82,33 +85,142 @@ def rpm_cmp_versions(v1, v2):
...
+            try:
+                version_file = os.path.join(repo_path, _file)
+                LOG.debug("%s: Reading version from %s using '%s'" %
+                          (self.pkg, version_file, regex))
+
+                with open(version_file, 'r') as f:
+                    match = re.search(regex, f.read())
+                self._repo_version = (
+                    '.'.join(s.strip() for s in match.groups() if s) or
+                    match.group().strip())
```
ghost: You don't need the "or" if the regex has parenthesis.

upgrade_versions.py
```diff
@@ -82,33 +85,142 @@ def rpm_cmp_versions(v1, v2):
...
+                                                         self._repo_version))
+                break
+            except:
+                pass
+        else:
+            LOG.error("%s: Could not find version in the source repo." %
+                      self.pkg)
+            msg = "Could not find %s's version in the source repo." % self.pkg
+            if not self.pkg.version_file_regex:
+                msg += " Try specifying version_file_regex in the .yaml file."
```
ghost: `" Try specifying version file and regex in the .yaml file."`
They are separate variables in the yaml file.

upgrade_versions.py
```diff
@@ -82,33 +85,142 @@ def rpm_cmp_versions(v1, v2):
...
+
+    def _read_spec(self):
+        self._spec_version = rpm_query_spec_file('version', self.pkg.specfile)
+        LOG.info("%s: Current version: %s" % (self.pkg, self._spec_version))
+        self._spec_release = rpm_query_spec_file(
+            'release', self.pkg.specfile).split('.')[0]
+
+    def _read_version_from_repo(self, repo_path):
+
+        version_regexes = [
```
ghost: I'd name this `version_files_and_regexes`

upgrade_versions.py
```diff
@@ -82,33 +85,142 @@ def rpm_cmp_versions(v1, v2):
...
+    def release(self):
+        return self._spec_release
+
+    def update(self):
+        changelog = None
+
+        pkg = copy.copy(self.pkg)
+        pkg.commit_id = None
+        pkg.download_source_code()
+        pkg.commit_id = pkg.repository.repo.head.target.hex[:7]
```
ghost, line 108: If commit_id hasn't changed, you don't need to do anything:

```
if pkg.commit_id != self.pkg.commit_id:
  LOG.debug("No changes for package %s", pkg.name)
  return
```

ghost (changes requested):

upgrade_versions.py
farosas: That was on purpose because I'm using `str.strip()` later on, but I guess there's no harm in changing it.

upgrade_versions.py
farosas: Yes. I was again trying to keep thinks flexible but perhaps there's no real benefit in not using groups.

upgrade_versions.py
farosas: Yep, and the commit message is wrong too.

ghost (approved):

maurorodrigues (approved):

fabianorosas:
skip ci

farosas:
start tests

maurorodrigues:
@fabianorosas Does this depends on something other than open-power-host-os/versions#43 right now??

farosas:
> @fabianorosas Does this depends on something other than open-power-host-os/versions#43 right now??

~~No, it doesn't.~~ I have one more change about cloning the versions repo ready to send but it is an improvement on what we already have today. It also depends on this PR so it would be good to merge it first.

EDIT: Actually, it does depend on https://github.com/open-power-host-os/versions/pull/47 because it needs the updated kernel.spec.

PR #99, opened 2016-10-20
Improve log handling

farosas:
Removes the creation of duplicated log entries
Adds a Formatter to the log handler
Replaces print statements with log.info

### DIFF
--- a/lib/log_helper.py
+++ b/lib/log_helper.py
@@ -20,21 +20,32 @@ import sys
 
 
 class LogHelper(object):
-    def __init__(self, logfile=None, level=logging.DEBUG, verbose=False):
+    def __init__(self, logfile=None, verbose=False):
         self.logfile = logfile
-        self.loglevel = level
 
         self._directory_setup()
 
         logger = logging.getLogger()
-        logging.basicConfig(filename=logfile, level=level)
-        print("Logs available at %s" % logfile)
+        logger.setLevel(logging.DEBUG)
+
+        sh = logging.StreamHandler(sys.stdout)
+        if verbose:
+            sh.setLevel(logging.DEBUG)
+        else:
+            sh.setLevel(logging.INFO)
+        sh.setFormatter(logging.Formatter('%(message)s'))
+        logger.addHandler(sh)
+
+        logger.info("Logs available at %s" % logfile)
+
         # NOTE(maurosr): RotatingFileHandler expects file size in bytes, in
         # short terms we're defining 2MB limit here.
-        logger.addHandler(logging.handlers.RotatingFileHandler(
-            logfile, maxBytes=2 << 20, backupCount=1))
-        if verbose:
-            logger.addHandler(logging.StreamHandler(sys.stdout))
+        rfh = logging.handlers.RotatingFileHandler(logfile, maxBytes=2 << 20,
+                                                   backupCount=1)
+        rfh.setLevel(logging.DEBUG)
+        rfh.setFormatter(logging.Formatter('%(asctime)s | %(levelname)s | '
+                                           '%(name)s: %(message)s'))
+        logger.addHandler(rfh)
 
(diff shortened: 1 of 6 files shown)


maurorodrigues:

farosas:
FYI, this increases the default stdout log level from no logging + some print messages to all the INFO messages. @olavphibm and I believe this to be a good thing. If anyone thinks we should allow stdout logging to be silenced completely let me know and I'll update the patch.

maurorodrigues:
test_plz

maurorodrigues:
good stuff, -1 for silence.. let's just wait the tests

lib/log_helper.py
```diff
@@ -20,21 +20,34 @@
...
+        rfh = logging.handlers.RotatingFileHandler(logfile, maxBytes=2 << 20,
+                                                   backupCount=1)
+        rfh.setLevel(logging.DEBUG)
+        rfh.setFormatter(logging.Formatter('%(asctime)s | %(levelname)s | '
+                                           '%(name)s: %(message)s'))
+        logger.addHandler(rfh)
+
         if verbose:
-            logger.addHandler(logging.StreamHandler(sys.stdout))
+            logger.handlers[0].setLevel(logging.DEBUG)
```
maurorodrigues: please move the whole if verbose to line 33.. something like

```
if verbose:
    sh.setLevel(logging,DEBUG)
else:
    sh.setLevel(logging.INFO)
```

Just to avoid issues with a change in the order of the handlers (aka avoid to use index like logger.handlers[0])

lib/log_helper.py
```diff
@@ -20,21 +20,34 @@
...
+        sh.setFormatter(logging.Formatter('%(message)s'))
+        logger.addHandler(sh)
+
+        logger.info("Logs available at %s" % logfile)
+
         # NOTE(maurosr): RotatingFileHandler expects file size in bytes, in
         # short terms we're defining 2MB limit here.
-        logger.addHandler(logging.handlers.RotatingFileHandler(
-            logfile, maxBytes=2 << 20, backupCount=1))
+        rfh = logging.handlers.RotatingFileHandler(logfile, maxBytes=2 << 20,
```
maurorodrigues: Suggestion only: maxBytes could be configurable huh?

maurorodrigues (changes requested):
couple of small changes

lib/log_helper.py
ghost: That would be nice. Should we add TODO or IMPROVEMENT tags for this kind of stuff?

lib/log_helper.py
ghost: I agree.

ghost (changes requested):
I don't see any benefit from a silent/quiet option right now.

lib/log_helper.py
maurorodrigues: yep, sounds good to me

lib/log_helper.py
```diff
@@ -20,21 +20,35 @@
...
+        logger.info("Logs available at %s" % logfile)
+
         # NOTE(maurosr): RotatingFileHandler expects file size in bytes, in
         # short terms we're defining 2MB limit here.
-        logger.addHandler(logging.handlers.RotatingFileHandler(
-            logfile, maxBytes=2 << 20, backupCount=1))
-        if verbose:
-            logger.addHandler(logging.StreamHandler(sys.stdout))
+        if not rotate_size:
+            rotate_size = 2 << 20
```
ghost, line 44: I think its better to set the constant directly as a default to the parameter.

farosas: I don't want to lose the comment from line 41 and I think moving it up there would be strange.

ghost (approved):

maurorodrigues (approved):

Issue #69, opened 2016-09-16
Allow dependency packages to be built standalone


olavphibm:
When a package depends on another, the latter must be located in the dependencies directory and must be mentioned in the former's yaml file, so that it is used in the build process.
Placing a package in the dependency directory disallows it to be built standalone.

olavphibm:
Closed by #109

Issue #57, opened 2016-09-05
Take a look at gitpython module


maurorodrigues:
Pygit2 is pretty cool but released version (at centOS's yum repo) is not that complete. Let's take a look at gitpython and some other modules to deal with git.

olavphibm:
Fixed by #105

Issue #46, opened 2016-09-01
Improve overall build automation


maurorodrigues:
- Submit changes made by update scripts in the projects commit id to github

rmatinata-ibm:
This is about how do we push the changes back to github....

olavphibm:
Fixed by #105

Issue #43, opened 2016-09-01
Create release notes process


maurorodrigues:
Let's use github.io pages to publish it.
Such page would be used to publish externally build's release notes (maybe a table of functionalities) and tests results/status.

cc/ @sathnaga @rquander FYI

rmatinata-ibm:
Release Notes should contain:
 - Changelog
 - SHA1s
 - Features added
 - Known Issues
 - Closed Issues

maurorodrigues:
Let's try a best effort to start having this at september, 14th

mopsfelder:
The https://open-power-host-os.github.io/ is already in place.

Thank you all!

PR #105, opened 2016-11-07
Weekly builds automation

ghost:
General code refactoring.
Separate BuildManager and Package classes RPM specific code.
Remove dependencies and build_dependencies sub directories from versions repo. See open-power-host-os/versions#51.
Replace pygit2 library with gitpython.
Automatically upgrade package versions and push to versions repository.
Automatically create release notes base on versions repository and push to open-power-host-os.github.io repository.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,15 +33,15 @@ $ sudo yum localinstall epel-release-7-5.noarch.rpm # Note the version may chang
 
  - bzip2
  - git
+ - GitPython
  - mock
- - python-pygit2
  - PyYAML
  - svn
  - wget
 
 ::
 
-$ sudo yum install -y bzip2 git mock python-pygit2 PyYAML svn wget
+$ sudo yum install -y bzip2 git GitPython mock PyYAML svn wget
 
 Settings
 --------
--- a/host-os-build.py
+++ b/host-os-build.py
@@ -34,6 +34,7 @@ def main(args):
     packages_to_build = (CONF.get('default').get('packages')
                          or config.discover_software())
 
+    LOG.info("Building packages: %s", ", ".join(packages_to_build))
     build_manager = manager.BuildManager(packages_to_build)
     return build_manager()
 
--- a/host-os-build.py
+++ b/host-os-build.py
@@ -29,13 +29,12 @@ LOG = logging.getLogger(__name__)
 
 
 def main(args):
-    conf = utils.setup_default_config()
-    utils.setup_versions_repository(conf)
-    conf['default']['packages'] = conf['default']['packages'] if (
-        conf.get('default').get('packages')) else (
-            config.discover_software())
+    CONF = utils.setup_default_config()
+    utils.setup_versions_repository(CONF)
+    packages_to_build = (CONF.get('default').get('packages')
+                         or config.discover_software())
 
-    build_manager = manager.BuildManager()
+    build_manager = manager.BuildManager(packages_to_build)
     return build_manager()
 
 if __name__ == '__main__':
--- a/host-os-build.py
+++ b/host-os-build.py
@@ -41,7 +41,13 @@ def main(args):
 
     LOG.info("Building packages: %s", ", ".join(packages_to_build))
     build_manager = manager.BuildManager(packages_to_build, distro)
-    return build_manager()
+    try:
+        build_manager()
+    except exception.BaseException as exc:
+        LOG.exception("Failed to build packages")
+        return exc.errno
+    else:
+        return 0
 
 if __name__ == '__main__':
     if os.getuid() is 0:
--- a/host-os-build.py
+++ b/host-os-build.py
@@ -20,7 +20,7 @@ from lib import config
 from lib import distro_utils
 from lib import exception
 from lib import log_helper
-from lib import manager
+from lib import build_manager
 from lib import repository
 from lib import utils
 
@@ -40,14 +40,15 @@ def main(args):
         CONF.get('default').get('arch_and_endianness'))
 
     LOG.info("Building packages: %s", ", ".join(packages_to_build))
-    build_manager = manager.BuildManager(packages_to_build, distro)
+    bm = build_manager.BuildManager(packages_to_build, distro)
     try:
-        build_manager()
+        bm()
     except exception.BaseException as exc:
         LOG.exception("Failed to build packages")
         return exc.errno
     else:
         return 0
+
 
 if __name__ == '__main__':
     if os.getuid() is 0:
--- a/host-os-build.py
+++ b/host-os-build.py
@@ -17,6 +17,7 @@ import os
 import sys
 
 from lib import config
+from lib import distro_utils
 from lib import exception
 from lib import log_helper
 from lib import manager
@@ -33,9 +34,13 @@ def main(args):
     utils.setup_versions_repository(CONF)
     packages_to_build = (CONF.get('default').get('packages')
                          or config.discover_software())
+    distro = distro_utils.get_distro(
+        CONF.get('default').get('distro_name'),
+        CONF.get('default').get('distro_version'),
+        CONF.get('default').get('arch_and_endianness'))
 
     LOG.info("Building packages: %s", ", ".join(packages_to_build))
-    build_manager = manager.BuildManager(packages_to_build)
+    build_manager = manager.BuildManager(packages_to_build, distro)
     return build_manager()
 
 if __name__ == '__main__':
--- a/host-os-build.py
+++ b/host-os-build.py
@@ -39,7 +39,8 @@ def main(args):
                          rotate_size=conf.get('default').get('log_size'))
     try:
         # setup versions directory
-        path, dirname = os.path.split(config.COMPONENTS_DIRECTORY)
+        path, dirname = os.path.split(
+            conf.get('default').get('build_versions_repo_dir'))
         repository.Repo(
             dirname, conf.get('default').get('build_versions_repository_url'),
             path, conf.get('default').get('build_version'))
--- a/host-os-build.py
+++ b/host-os-build.py
@@ -21,6 +21,7 @@ from lib import exception
 from lib import log_helper
 from lib import manager
 from lib import repository
+from lib import utils
 
 import pygit2
 
@@ -28,30 +29,12 @@ LOG = logging.getLogger(__name__)
 
 
 def main(args):
-    try:
-        conf = config.get_config().CONF
-    except OSError:
-        print("Failed to parse settings")
-        return 2
+    conf = utils.setup_default_config()
+    utils.setup_versions_repository(conf)
+    conf['default']['packages'] = conf['default']['packages'] if (
+        conf.get('default').get('packages')) else (
+            config.discover_software())
 
-    log_helper.LogHelper(logfile=conf.get('default').get('log_file'),
-                         verbose=conf.get('default').get('verbose'),
-                         rotate_size=conf.get('default').get('log_size'))
-    try:
-        # setup versions directory
-        path, dirname = os.path.split(
-            conf.get('default').get('build_versions_repo_dir'))
-        repository.Repo(
-            dirname, conf.get('default').get('build_versions_repository_url'),
-            path, conf.get('default').get('build_version'))
-
-        conf['default']['packages'] = conf['default']['packages'] if (
-            conf.get('default').get('packages')) else (
-                config.discover_software())
-
-    except exception.RepositoryError as e:
-        LOG.exception("Script failed")
-        return e.errno
     build_manager = manager.BuildManager()
     return build_manager()
 
--- a/host-os-build.py
+++ b/host-os-build.py
@@ -24,8 +24,6 @@ from lib import build_manager
 from lib import repository
 from lib import utils
 
-import pygit2
-
 LOG = logging.getLogger(__name__)
 
 
--- a/lib/build_manager.py
+++ b/lib/build_manager.py
@@ -18,7 +18,7 @@ import logging
 import lib.centos
 from lib import exception
 import lib.scheduler
-from lib import utils
+from lib.packages_manager import PackagesManager
 from lib.rpm_package import RPM_Package
 
 LOG = logging.getLogger(__name__)
@@ -26,15 +26,14 @@ LOG = logging.getLogger(__name__)
 
 class BuildManager(object):
     def __init__(self, packages_names, distro):
-        self.packages_names = packages_names
-        self.packages = []
+        self.packages_manager = PackagesManager(packages_names)
         self.distro = distro
         self.repositories = None
 
     def __call__(self):
         try:
-            self.prepare_packages()
-
+            self.packages_manager.prepare_packages(
+                packages_class=RPM_Package, distro=self.distro)
         # distro related issues
         except (exception.DistributionNotSupportedError,
                 exception.DistributionVersionNotSupportedError,
@@ -42,21 +41,9 @@ class BuildManager(object):
             LOG.error("Error during distribution detection. "
                       "See the logs for more information")
             raise
-        # package issues
-        except exception.PackageError:
-            LOG.error("Failed to load the package in components. "
-                      "See the logs for more information")
-            raise
 
         self.build()
 
     def build(self):
         scheduler = lib.scheduler.Scheduler()
-        self.distro.build_packages(scheduler(self.packages))
-
-    def prepare_packages(self, download_source_code=True):
-        for package_name in self.packages_names:
-            package = RPM_Package.get_instance(package_name, self.distro)
-            if download_source_code:
-                package.download_files()
-            self.packages.append(package)
+        self.distro.build_packages(scheduler(self.packages_manager.packages))
--- a/lib/config.py
+++ b/lib/config.py
@@ -21,7 +21,6 @@ import sys
 import yaml
 
 LOG = logging.getLogger(__name__)
-COMPONENTS_DIRECTORY = os.path.join(os.getcwd(), "components")
 
 config_parser = None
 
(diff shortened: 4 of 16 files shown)


ghost:
This fixes #69, #57, #46 and #43 (partially).

lib/repository.py
```diff
@@ -17,103 +17,83 @@
...
+import gitdb
 
 from lib import exception
 
 LOG = logging.getLogger(__name__)
 
+def get_git_repository(name, remote_repo_url, parent_dir_path):
+    """
+    Get a local git repository with specified name located in a
+    subdirectory of the parent dir.
```
farosas: The structure is `parent_dir_path/subdirectory/name` or `parent_dir_path/name`? Code implies the latter.

lib/repository.py
```diff
@@ -124,25 +104,23 @@ def _update_submodules(self):
         Update repository submodules, initializing them if needed.
         """
         cmd = "git submodule init; git submodule update"
```
farosas: AFAIK, gitpython supports this. Something like:

```
for submodule in self.submodules:
    submodule.update(init=True)
```

lib/package.py
```diff
@@ -64,12 +64,6 @@ def __init__(self, name, distro, download=True):
         self.package_file = os.path.join(self.package_dir, self.name,
                                          '%s.yaml' % self.name)
 
-        #TODO(maurosr): Improve this piece of code, actions shouldn't go in
-        # __init__, let's refactor in order to move the download action.
-        self.load_package(name, distro)
```
farosas: I think the loading of the yaml file should be part of the package initialization since we can't really do anything without having read the package's .yaml file.

What about leaving the call to `load` here and:
- move the call to `self._download_build_files` (l 129) and `self.download_source_code()` to a single method (say, 'download_sources').
- only load packages' .yaml files during `load` and thus, move the calls to 'download_sources' of each dependency to 'download_sources' itself:

```
def __init__():
    self.load()
def download_sources():
    <downloads source code>
    <downloads build files>
    for dep in self.dependencies:
        dep.download_sources()
```

That way we might even think about stopping propagating the `download` flag all over the code and leave it in the caller:

```
for package_name in self.packages_names:
    package = package.Package.get_instance(package_name, self.distro)
    if download_source_code:
        package.download_sources()
    self.packages.append(package)
```

README.rst
```diff
@@ -33,15 +33,15 @@ $ sudo yum localinstall epel-release-7-5.noarch.rpm # Note the version may chang
...
  - mock
- - python-pygit2
  - PyYAML
  - svn
  - wget
 
 ::
 
-$ sudo yum install -y bzip2 git mock python-pygit2 PyYAML svn wget
+$ sudo yum install -y bzip2 git GitPython mock PyYAML svn wget
```
maurorodrigues, line 44: let's remember to emphasize this in the release notes

lib/repository.py
```diff
@@ -17,103 +17,83 @@
...
+        LOG.info("Found existent repository at destination path %s" % repo_path)
+        return GitRepository(repo_path)
+    else:
+        LOG.info("Cloning into %s" % repo_path)
+        try:
+            return GitRepository.clone_from(remote_repo_url, repo_path)
+        except git.exc.GitCommandError:
+            message = "Failed to clone repository"
+            LOG.exception(message)
+            raise exception.RepositoryError(message=message)
```
maurorodrigues, line 63: I feel like this could go within the GitRepository class, don't you think?

lib/repository.py
```diff
@@ -17,103 +17,83 @@
...
 
+def get_git_repository(name, remote_repo_url, parent_dir_path):
+    """
+    Get a local git repository with specified name located in a
+    subdirectory of the parent dir.
+    If it does not exist, clone it from the specified URL.
+    """
+    repo_path = os.path.join(parent_dir_path, name)
+    if os.path.exists(repo_path):
+        LOG.info("Found existent repository at destination path %s" % repo_path)
```
maurorodrigues: we can't confirm that we found the existent repository at the destination path before we instantiate it (which goes in the same direction of a comment bellow where I suggest to get this within GitRepository class)

lib/repository.py
```diff
@@ -17,103 +17,83 @@
...
-                LOG.info("Failed to fetch %s remote for %s"
-                         % (remote.name, self.repo_name))
+            except git.exc.GitCommandError:
+                LOG.debug("Failed to fetch %s remote for %s"
+                          % (remote.name, self.name))
                 pass
             else:
-                LOG.info("%(repo_name)s Repository updated" % vars(self))
+                LOG.info("Fetched changes for %s" % remote.name)
+            remote.fetch()
```
maurorodrigues: why are you calling remote.fetch() here and in L61?

upgrade_versions.py
```diff
@@ -217,6 +220,14 @@ def main(args):
         CONF.get('default').get('distro_name'),
         CONF.get('default').get('distro_version'),
         CONF.get('default').get('arch_and_endianness'))
+    committer_name = CONF.get('default').get('committer_name')
+    committer_email = CONF.get('default').get('committer_email')
```
farosas, line 257: Why don't you move these directly into rpm_bump_spec?

lib/config.py
```diff
@@ -129,10 +129,18 @@ def _setup_config_parser_options(self):
...
+                                 'pushing',
+                                 default='master')
         self.parser.add_argument('--committer-name',
-                                 help='Name used when bumping spec files')
+                                 help='Name used when creating a commit and '
+                                 'bumping spec files')
         self.parser.add_argument('--committer-email',
-                                 help='Email used when bumping spec files')
+                                 help='Email used when creating a commit and '
+                                 'bumping spec files')
```
farosas, line 143: We should think about splitting this parser into two.

lib/repository.py
maurorodrigues: Note: of course, we can make use a similar structure in the future when (if) we start supporting different VCS by turning this into a factory in an abstract Repository class

upgrade_versions.py
```diff
@@ -212,18 +213,49 @@ def _read_version_from_repo(self, repo_path):
...
+    versions_repo.index.add(["*"])
+
+    LOG.info("Committing changes to local repository")
+    commit_message = "Weekly build {date}".format(date=release_date)
+    actor = git.Actor(committer_name, committer_email)
+    versions_repo.index.commit(commit_message, author=actor, committer=actor)
+
+    LOG.info("Pushing changes to remote repository")
+    remote = versions_repo.remote(VERSIONS_REPO_REMOTE)
+    refspec = "HEAD:refs/heads/{}".format(versions_repo_push_branch)
```
farosas, line 239: Just checking, but we don't have this requirement currently do we?

upgrade_versions.py
```diff
@@ -241,6 +273,10 @@ def main(args):
             LOG.exception("Failed to update versions")
             return e.errno
 
+    release_date = datetime.today().date().isoformat()
+    push_new_versions(versions_repo, release_date, push_repo_url,
+                      push_repo_branch, committer_name, committer_email)
```
farosas, line 282: What happens if no package is updated?

lib/repository.py
ghost: The latter. Clarifying doc string.

lib/repository.py
ghost: Yep. I'll add a TODO and do it in another patch, along with archive.
This patch should be only about removing pygit2.

lib/package.py
```diff
@@ -26,8 +26,6 @@
 
 CONF = config.get_config().CONF
 LOG = logging.getLogger(__name__)
-BUILD_DEPENDENCIES = "build_dependencies"
-DEPENDENCIES = "dependencies"
```
maurorodrigues, line 29: Any chance we can get this whole commit into a separated PR?

lib/package.py
maurorodrigues: > I think the loading of the yaml file should be part of the package initialization since we can't really do anything without having read the package's .yaml file.

Agreed, it isn't really an action in the sense of an object's action, it's an inner action necessary for it's existence.

> What about leaving the call to load here and:

move the call to self._download_build_files (l 129) and self.download_source_code() to a single method (say, 'download_sources').
only load packages' .yaml files during load and thus, move the calls to 'download_sources' of each dependency to 'download_sources' itself:

+1 but it's totally fine to do it in a separate patch series

lib/rpm_package.py
```diff
@@ -0,0 +1,86 @@
...
+        Read yaml file describing this package.
+        Optionally, download its source code.
+        """
+        super(RPM_Package, self).load()
+        try:
+            with open(self.package_file, 'r') as package_file:
+                package_info = yaml.load(package_file).get('Package')
+        except IOError:
+            raise exception.PackageDescriptorError(
+                "Failed to open %s's YAML descriptor" % self.name)
```
maurorodrigues: I would prefer to create a property named `distribution_data` (or something like that, this name isn't a great one)  in the superclass and load everything from it, instead of reading the `self.package_file` again here.

lib/build_manager.py
```diff
@@ -18,13 +18,13 @@
...
 import lib.scheduler
-from lib import utils
+from lib.packages_manager import PackagesManager
 from lib.rpm_package import RPM_Package
 
 LOG = logging.getLogger(__name__)
 
 
-class BuildManager(object):
+class BuildManager(PackagesManager):
```
maurorodrigues: split makes perfect sense but the Inheritance here seems wrong to me, there isn't a inheritance situation here, I mean I don't see `BuildManager` being a specialization of `PackagesManager` (in fact I think if inheritance should be used should be the other way around, `PackagesManager` inheriting from `BuildManager`).

A composition seem more appropriate, ie, an instance of `PackagesManager` should be an attribute here (like `PackageManager` is one tool for the whole Build process controlled by the `BuildManager` which eventually will have another tool to build something else (like an iso).

lib/build_manager.py
```diff
@@ -33,20 +33,17 @@ def __init__(self, packages_names, distro):
...
-        # package issues
-        except exception.PackageError as exc:
-            LOG.exception("Failed to load the package in components. "
-                          "See the logs for more information")
-            return exc.errno
+        else:
+            if error_number:
+                return error_number
 
         return self.build()
```
maurorodrigues, line 46: I think both methods below (which github doesn't allow me to comment it seems) belong to PackagesManager, don't you think? then the `build` method here would be just calling all its manager's build process (see the composition suggestion above and the idea of having multiple build artifacts)

upgrade_versions.py
```diff
@@ -210,20 +209,10 @@ def _read_version_from_repo(self, repo_path):
...
-        # setup versions directory
-        path, dirname = os.path.split(
-            CONF.get('default').get('build_versions_repo_dir'))
-        repository.Repo(
-            dirname, CONF.get('default').get('build_versions_repository_url'),
-            path, CONF.get('default').get('build_version'))
-    except exception.RepositoryError as exc:
-        LOG.exception("Failed to checkout versions repository")
-        return exc.errno
+    LOG.info("Updating packages version...")
```
maurorodrigues: This log.info makes more sense within `setup_versions_repository` and Updated instead of Updating (like it is a step already done)

lib/manager.py
```diff
@@ -31,8 +29,8 @@
 
 
 class BuildManager(object):
-    def __init__(self, packages=None):
-        self.packages_list = packages or CONF.get('default').get('packages')
+    def __init__(self, packages_names):
+        self.packages_names = packages_names
```
maurorodrigues: I don't see how packages_names is better than packages_list for a list of packages (I mean it's a list, not a string with comma or space separated names), but I'll not give you a hard time about it either

upgrade_versions.py
```diff
@@ -213,7 +213,7 @@ def main(args):
     utils.setup_versions_repository(CONF)
     packages_to_update = CONF.get('default').get('packages') or PACKAGES
 
-    LOG.info("Updating packages version...")
+    LOG.info("Updating packages versions: %s", ", ".join(packages_to_update))
```
maurorodrigues: this may give the wrong impression that all packages are being updated (it's something that we only know later in the process though)

host-os-build.py
```diff
@@ -34,6 +34,7 @@ def main(args):
     packages_to_build = (CONF.get('default').get('packages')
                          or config.discover_software())
 
+    LOG.info("Building packages: %s", ", ".join(packages_to_build))
```
maurorodrigues, line 40: both this and the Log.info below don't bring anything to the table imho, scheduler does show pretty much the same info and ordered

lib/package.py
ghost: That's much better, thanks!

upgrade_versions.py
ghost: The workspace of our build machine is volatile and can be deleted after builds.
This pushes the changes to somewhere accessible (the bot's github fork, for example) to ease a possible pull request and merge into the main repository.

lib/config.py
maurorodrigues: huge +1... I'll create a task about it... Pretty sure I started it one day but then I realized that refactoring was need to decouple the settings from the host-os.py

upgrade_versions.py
ghost: This is to make sure the configs are set _before_ actually executing the commands. I think we should find a way of checking all the configs required to execute a given script.
I also don't know if it's a good idea to get the configs directly in the place they are used. This entangles the methods and classes with the config module.
One more point: those configs will also be passed to the function push_new_versions in the next patch.

upgrade_versions.py
maurorodrigues: good point.

upgrade_versions.py
```diff
@@ -239,6 +239,8 @@ def push_new_versions(versions_repo, release_date, versions_repo_push_url,
     refspec = "HEAD:refs/heads/{}".format(versions_repo_push_branch)
     push_info = remote.push(refspec=refspec)[0]
     LOG.debug("Push result: {}".format(push_info.summary))
+    if git.PushInfo.ERROR & push_info.flags:
+        raise repository.PushError(push_info)
```
maurorodrigues, line 244: This could be sqaushed with @35b0a414f53e5d99ceebfc14bfb96836a3dbcbfa but I'll not make a hard time about this either, we'll get into the meaning of atomicity and such all over again.

lib/config.py
ghost: Yeah, it needs some refactoring. There are configs which are required by some scripts and not used in others.

upgrade_versions.py
ghost: An empty commit will be created and pushed. Harmless, but not ideal.
I believe the Version.update method would need to be modified to return the update status.

maurorodrigues (changes requested):
Some points inline.

Note: I don't intend to be pedantic here and I do understand some of this refactoring is necessary to get this done but in general some commits here could go in separated PR for some reasons, for instance:
-  it's excessive complexity/discussion necessary (like the one which removes build_deps and deps hierarchy which I do remember agreed to get it done in this series even without the backwards compatibility piece)
  - simplicity (the typos or add log information)
  - relationship with the general thematic of the patch set (minimize try blocks for instance)

Let's keep this series as is, but also bear in mind that as it's good to split a feature in multiple patch it is as good to have PRs with a sole purpose.

maurorodrigues:
Also, from #43 would be nice if we can include this info as well: 
Known Issues (like the ivtv-firmware information) maybe a separated file, maybe even a separated task to add this if you prefer. 

Closed Issues (this one I think it's a separated task, may require some work with github api if people are not leaving this kind of information in the commit messages)

lib/repository.py
ghost: This should not be here.

lib/repository.py
ghost: That's right. I'm creating a GitRepository **init** method and placing this there.

lib/repository.py
ghost: Yes, that's better.

lib/package.py
ghost: Yes. I did this to be able to include dependencies in release notes, but that's not necessary.

lib/rpm_package.py
ghost: That's better indeed.
I've renamed package_info to package_data and added it as an object attribute (self.package_data).

lib/manager.py
farosas: This is really minor, but here is my 2 cents: I generally think that having the data type present in the variable name will eventually confuse someone when things change down the line. Also, one might be inclined to, e.g.: `packages_list[0].name`, after all we have a 'Package' class somewhere and this is a package list.

lib/build_manager.py
ghost: That makes sense.

lib/build_manager.py
ghost: I'm not sure which two methods you're refering to (**call** and prepare_packages?).
I've removed the method prepare_packages from BuildManager and had to add an argument "package_class" in PackagesManager.prepare_packages, to allow creating RPM packages.

One more think that I think could be changed is the **call** method. It's kind of confusing to call BuildManager()() and calling a named method like "build" or "prepare_packages" is more straightfoward. That can be done later, if you agree.

upgrade_versions.py
ghost: Right. Rephrased it.

upgrade_versions.py
ghost: This is only separated because I've found about it later and so I just created another commit. Squashing.

upgrade_versions.py
ghost: This is intended to mean the packages versions will be updated.
I've changed the message like you suggested in the other comment and there is already a log message when a repository is cloned/checked out.

host-os-build.py
ghost: I think they are useful at this point because the scheduler is only created after loading all the packages. Since we currently download sources during packages loading and that may take a long time, it is good to know beforehand if we passed the correct packages as parameters.
Besides, another patch replaces BuildManager with PackagesManager in the upgrade_versions script, and it does not use the scheduler.

upgrade_versions.py
maurorodrigues: this will be pushed to a versions fork into our bot's account? if so, let's keep it then since no harm is made unless we submit it to a PR

lib/package.py
maurorodrigues: cool thanks, it's just cause I would like to put more thought on the backwards compatibility matter, but I don't want to block this series either

lib/build_manager.py
maurorodrigues: ops sorry, the way github shows it in the conversation tab sucks for commenting in things that were not changed.
I was refering to `build` and `prepare_packages` (I think you can visualize them in this link https://github.com/open-power-host-os/builds/pull/105/commits/a3fc76ab4b8176c6688e94167fea7957f77c3a29#diff-d4f29a8a05d49edd09eeffae2d56abe5R54) 

> One more think that I think could be changed is the call method. It's kind of confusing to call BuildManager()() and calling a named method like "build" or "prepare_packages" is more straightfoward. That can be done later, if you agree.

+1

upgrade_versions.py
ghost: Yes, the PR still has to be created manually.

olavphibm:
test_plz

olavphibm:
test this please

olavphibm:
test this please

ghost:
@maurorodrigues 
I'll leave the adition of more info from #43 to other pull requests.

maurorodrigues (approved):

Issue #50, opened 2016-09-01
Customize/Remaster CentOS 7.2 ISO w/ hostOS built packages


maurorodrigues:

gutoyr:
Implemented by #110

PR #110, opened 2016-11-14
Add code to create a HostOS ISO image based on CentOS packages

gyribeiro:
Refactor scripts that start actions, now there is a single script that supports subcommands that handle all required actions.
Support to create Host OS ISO image was added, the subcommand to perform this action was added, as well.


lib/iso_spinner.py
```diff
@@ -0,0 +1,128 @@
...
+        LOG.debug("Copying rpm to spin repo directory")
+        rpm_dir = self.config.get('rpm_dir')
+        rpm_files = glob.glob(os.path.join(rpm_dir, "*.rpm"))
+        self._run_mock_command("--copyin %s %s" %
+                               (" ".join(rpm_files), mock_spin_repo_dir))
+
+        LOG.debug("Creating comps.xml")
+        # TODO: create comps.xml
+        comps_xml_file = "hostos-comps.xml"
+        self._run_mock_command("--copyin %s /" % comps_xml_file)
```
ghost: I'd write something like `"--copyin %s %s" % (comps_xml_file, comps_xml_chroot_path)` to make it clearer what the second parameter is and to make sure the createrepo command matches whenever someone modifies this.

lib/iso_spinner.py
```diff
@@ -0,0 +1,128 @@
...
+
+LOG = logging.getLogger(__name__)
+
+
+class MockPungiSpinner(object):
+
+    def __init__(self, config):
+        self.config = config
+        (self.distro, self.version, self.arch) = distro_utils.detect_distribution()
+        # only the major number must be consider
```
farosas: considered

lib/iso_spinner.py
```diff
@@ -0,0 +1,128 @@
...
+        self.version = re.sub("\..*$", "", self.version)
```
farosas: I'd go with `self.version.split('.')[0]` which is a bit more readable.

lib/iso_spinner.py
```diff
@@ -0,0 +1,128 @@
...
+    def _create_spin_kickstart(self):
+        kickstart_file = self.config.get('kickstart_file')
+        LOG.info("Creating spin kickstart file %s" % kickstart_file)
+
+        repo_url = self.config.get('distro_repo_url')
+        mock_spin_repo_name = self.config.get('mock_spin_repo').get('name')
+        mock_spin_repo_dir = self.config.get('mock_spin_repo').get('dir')
+        repo_url[mock_spin_repo_name] = "file://%s/" % mock_spin_repo_dir
+        with open(kickstart_file, "a+") as f:
+            for name, url in repo_url.items():
```
ghost: I'd name this `repo_urls` if it contains multiple URLs.

lib/iso_spinner.py
```diff
@@ -0,0 +1,128 @@
...
+            package_group_list.append("@%s" %
+                                      self.config.get('hostos_package_group'))
+            for package_group in package_group_list:
+                f.write(package_group + "\n")
+
+        self._run_mock_command("--copyin %s /" % kickstart_file)
+
+    def _spin(self):
+        LOG.info("Spinning ISO")
+        spin_cmd = "pungi -c %s --nosource --nodebuginfo --name %s --ver %s" % \
```
ghost: I prefer to use only parentheses instead of `\` to wrap lines.

lib/iso_spinner.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def build(self):
+        LOG.info("Starting ISO spin process")
+        self._setup()
+        self._spin()
+        self._save()
+
+    def _setup(self):
+        LOG.info("Initializing a chroot")
+        self._run_mock_command("-init")
```
farosas: Shouldn't it be '--init'?

lib/iso_spinner.py
```diff
@@ -0,0 +1,128 @@
...
+
+class MockPungiSpinner(object):
+
+    def __init__(self, config):
+        self.config = config
+        (self.distro, self.version, self.arch) = distro_utils.detect_distribution()
+        # only the major number must be consider
+        self.version = re.sub("\..*$", "", self.version)
+
+    def _run_mock_command(self, cmd):
```
farosas, line 38: I might have missed this discussion, have we decided not to use the Mock class for this?

iso_config.yaml
```diff
@@ -0,0 +1,37 @@
+log_file: "/var/tmp/host-os-iso-creator.log"
+mock_config: "./extras/centOS/7.2/mock/spin-iso-epel-7-ppc64le.cfg"
+rpm_dir: "./results"
+kickstart_file: "host-os.ks"
+hostos_package_group: "virtualization-hypervisor"
+package_group_list:
```
mopsfelder: Do we need groups like desktop-debugging, dial-up, gnome-_, guest-desktop-agents, internet-_, multimedia, x11?

iso_config.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - "@virtualization-hypervisor"
+    - "@virtualization-tools"
+    - "@web-server"
+    - "@x11"
+mock_spin_repo:
+    name: "hostos"
+    dir: "/spin-repo"
+distro_repo_url:
+    base: "http://mirror.aus.stglabs.ibm.com/centos-altarch/7/os/ppc64le/"
+    updates: "http://mirror.aus.stglabs.ibm.com/centos-altarch/7/updates/ppc64le/"
```
mopsfelder: This is an internal mirror.  What about http://mirror.centos.org/altarch/7/os/ppc64le/ for 'base' and http://mirror.centos.org/altarch/7/updates/ppc64le/ for 'updates'?

iso_config.yaml
```diff
@@ -0,0 +1,37 @@
+log_file: "/var/tmp/host-os-iso-creator.log"
```
ghost: In config.yaml, we have a "default" node containing all the other nodes. I think we should a similar convention in both files. That does not necessarily need to have the "default" node, but changing may be hard if we need to be backwards compatible.
The other config has `log_file: "/var/log/host-os/builds.log"` and it uses that same path for the three scripts that we have. I think we should use the "/var/log/host-os" here too and (unrelated to this PR) maybe use different files for different scripts (upgrade_versions, create_release_notes).

iso_config.yaml
```diff
@@ -0,0 +1,37 @@
...
+mock_config: "./extras/centOS/7.2/mock/spin-iso-epel-7-ppc64le.cfg"
+rpm_dir: "./results"
```
ghost: This could be a common config between this and the host-os-build "result_dir" RPM. It would be easier to generate the RPMs and then create the ISO in sequence. Maybe we could merge this file into config.yaml and use a sub-node "mock_iso" to separate the specific configs?

host-os-iso-creator.py
```diff
@@ -0,0 +1,43 @@
+# Copyright (C) IBM Corp. 2016.
```
ghost: PEP8 is against using dashes in file names because they cannot be imported with a standard import statement. I suggest create_host_os_iso.py
I think we should rename the host-os-build.py file too, but I don't know if we can do this now with the backwards compatibility issue.

host-os-iso-creator.py
```diff
@@ -0,0 +1,43 @@
...
+                         verbose=True),
+    spinner = iso_spinner.MockPungiSpinner(config)
+    spinner.build()
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--config-file', '-c',
+                        help='Path of the configuration file for create iso',
+                        default='./iso_config.yaml')
```
ghost: We need to find an easier way to have different arguments in different scripts. I don't have any suggestions right now though, so that's fine.

iso_config.yaml
farosas: +1 for merging and having a 'sub-node'

host-os-iso-creator.py
ghost: About my backwards compatibility concerns, @maurorodrigues explained to me that this is a problem only to the versions repository, so they may be discarded here.
I keep my suggestion of naming to create_host_os_iso.py and (unrelated to this PR) create_host_os_build.py or create_host_os_packages.py

lib/iso_spinner.py
gyribeiro: Done

lib/iso_spinner.py
gyribeiro: Done

lib/iso_spinner.py
gyribeiro: It was decided to keep rpm package builder and ISO image creator codes in separate modules, since they do not share any code.

Just a quick note: Mock class, in my opinion, is a misleading name because it is not a wrapper for mock command as you may notice, it is just rpm packages builder that uses mock.

lib/iso_spinner.py
gyribeiro: Done

iso_config.yaml
gyribeiro: Short answer is not, the reason for this group list is that it came from official CentOS 7 ppc64le ISO, however, I've made it flexible exactly for this scenario, i.e., hostos decided to provide a custom CentOS ISO instead of the official ISO + hostos packages.

If the team prefers I can use a custom list instead of the official one. If this is the case we can discuss a minimum group list.

iso_config.yaml
gyribeiro: Agree, in the beginning it was using official CentOS mirror, but I've changed it during my tests, pointing them to an internal one to speed up the process and forgot to revert it.

lib/iso_spinner.py
gyribeiro: Done

lib/iso_spinner.py
gyribeiro: Done

lib/iso_spinner.py
gyribeiro: Done

iso_config.yaml
gyribeiro: I didn't get your point regarding having or not  a "default" node in all yaml files. Do you want to add this node in the ISO yaml, as well?

Regarding the log, I was aware that the package builder uses /var/log/host-os to store its log, however, at the time I've written this code I did realized that I should run setup_environment.py previously to grant write access for /var/log. I will change it to "log_file: /var/log/host-os/iso.log", is it works for you?

iso_config.yaml
gyribeiro: Initially, I've thought about it but as we decided to keep to rpm package builder and ISO creator code separate, I kept the same approach for the yaml file.

If the team agrees with this proposal, I don't see any problem in merger both yaml files as you suggest.

host-os-iso-creator.py
gyribeiro: Ok, rename done ("-_ -> "_"), I've just used dash to follow the repository pattern.

host-os-iso-creator.py
gyribeiro: Agreed. We can have a discussion about it to find a better approach to avoid duplicated code spread in the project.

iso_config.yaml
ghost: About the "default" node: I think we should follow the same format in both yaml file, but I don't really get why we have this "default" node, so up to you.
"/var/log/host-os/iso.log" is fine.

lib/iso_spinner.py
```diff
@@ -0,0 +1,131 @@
...
+        LOG.debug("Copying rpm to spin repo directory")
+        rpm_dir = self.config.get('rpm_dir')
+        rpm_files = glob.glob(os.path.join(rpm_dir, "*.rpm"))
+        self._run_mock_command("--copyin %s %s" %
+                               (" ".join(rpm_files), mock_spin_repo_dir))
+
+        LOG.debug("Creating comps.xml")
+        # TODO: create comps.xml
+        comps_xml_file = "hostos-comps.xml"
+        comps_xml_chroot_path = "/"
```
ghost: I was actually thinking about setting
`comps_xml_chroot_path = os.path.join("/", comps_xml_file)`

lib/iso_spinner.py
```diff
@@ -0,0 +1,131 @@
...
+        self._run_mock_command("--copyin %s %s" %
+                               (comps_xml_file, comps_xml_chroot_path))
+
+        LOG.debug("Creating spin repo")
+        createrepo_cmd = "createrepo -v -g /%s %s" % (comps_xml_file,
```
ghost: So we could use `comps_xml_chroot_path` here and make the link with the _run_mock_command parameters above more obvious to people unfamiliar with mock.

lib/iso_spinner.py
gyribeiro: done

iso_config.yaml
gyribeiro: For now, I will keep it without the "default" node, I will start a discussion about whether it is worth to merge the config files or not.

iso_config.yaml
gutoyr: I did some tests and was able to reduce the package list to a minimum set, which speeds up the process and reduces the ISO size, hence, I will keep your suggestion and leave a tiny set of packages.

iso_config.yaml
gyribeiro: I did some tests and was able to reduce the package list to a minimum set, which speeds up the process and reduces the ISO size, hence, I will keep your suggestion and leave a tiny set of packages.

lib/iso_spinner.py
```diff
@@ -0,0 +1,158 @@
...
+                "@%s" % self.config.get('hostos_package_group').get('name'))
+            if hostos_package_group_name not in package_group_list:
+                package_group_list.append(hostos_package_group_name)
+            for package_group in package_group_list:
+                f.write(package_group + "\n")
+
+            # FIXME: linux-firmware provided by HostOS conflicts with
+            #        ivtv-firmware, this hack must be removed as soon as its
+            #        problem has been sorted out
+            f.write("-ivtv-firmware\n")
```
ghost: Do we need this here if it was already excluded from the repos above?

lib/iso_spinner.py
```diff
@@ -0,0 +1,158 @@
...
+        LOG.info("Spinning ISO")
+        spin_cmd = ("pungi -c %s --nosource --nodebuginfo --name %s --ver %s" %
+                    (self.config.get('kickstart_file'),
+                     self.distro, self.version))
+        self._run_mock_command("--shell '%s'" % spin_cmd)
+
+    def _save(self):
+        iso_file = "%s-DVD-%s-%s.iso" % (self.distro, self.arch, self.version)
+        checksum_file = ("%s-%s-%s-CHECKSUM" %
+                         (self.distro, self.version, self.arch))
```
ghost, line 140: I dislike that the ISO and checksum file names are non-uniform. If you're just copying those formats from CentOS, then that's fine.

lib/iso_spinner.py
gyribeiro: agree, removed

lib/iso_spinner.py
gyribeiro: This is odd, unfortunately we can not change it because pungi (spin tool) adopts different file name patterns for ISO and checksum.

host_os.py
```diff
@@ -0,0 +1,38 @@
...
+    CONF = utils.setup_default_config()
+    subcommand = CONF.get('default').get('subcommand')
+
+    if os.getuid() is 0:
+        print("Please, do not run this script as root, run "
+              "setup_environment.py script in order to properly setup user and"
+              " directory for build scripts")
+        sys.exit(3)
+
+    sys.exit(SUBCOMMANDS.get(subcommand).run(CONF))
```
ghost: `get(subcommand)` will return None and raise an `AttributeError` if the key does not exist. I think it's more straightforward to use `[subcommand]`, so a `KeyError` will be raised.

gyribeiro: Done.

lib/config.py
```diff
@@ -98,56 +99,132 @@ def _setup_config_parser_options(self):
...
         self.parser.add_argument('--verbose', '-v',
                                  help='Set the scripts to be verbose',
                                  action='store_true')
-        self.parser.add_argument('--result-dir', '-r',
-                                 help='Directory to save the RPMs.',
-                                 default='./result')
-        self.parser.add_argument('--repositories-path', '-R',
-                                 help='Directory where to clone code repositories',
-                                 default='/var/lib/host-os/repositories')
         self.parser.add_argument('--keep-builddir',
```
farosas: Isn't this specific to build-package?

lib/config.py
```diff
@@ -98,56 +99,132 @@ def _setup_config_parser_options(self):
...
-        self.parser.add_argument('--committer-email',
-                                 help='Email used when creating a commit and '
-                                 'bumping spec files')
 
+        subparser = self.parser.add_subparsers(
+            dest="subcommand",
+            help="Available subcommands")
+
+        # create the parser for the "build-package" subcommand
+        parser_package = subparser.add_parser('build-package',
```
farosas: What if we had a list of argument groups like so:
```
build_repo = [
    dict(option='--build-versions-repository-url', help='Build versions repository URL')
    dict(option='--build-versions-repo-dir', help='Directory to clone the build versions repository', default='./components')
]
push_repo = [...]
```

And then made each command 'subscribe' to them:
```
SUBCOMMANDS = [
    ('build-package', 'Build packages', [build_repo] ),
    ('release-notes', 'Create release notes', [push_repo, buld_repo]),
    ...
]
```

```
def add_args():
    for command, help, arg_groups in SUBCOMMANDS:
        subparser = subparser.add_parser(command, help)
        for arg_group in arg_groups:    
            for arg in arg_group:
                subparser.add_argument(**arg)
```

lib/config.py
```diff
@@ -98,56 +99,132 @@ def _setup_config_parser_options(self):
...
+        self._add_setup_environment_args(parser_setup_environment)
+
+        # create the parser for the "build-iso" subcommand
+        parser_iso = subparser.add_parser('build-iso', help="Build ISO image")
+        self._add_iso_args(parser_iso)
+
+    def _add_build_repo_args(self, subparser):
+        """
+        Configures build repository arguments for subparser object to match the
+            expected configuration.
```
farosas: If my suggestion above is not taken, I think we should alter these docs to something more descriptive, like: "Args related to building the repo".

lib/config.py
```diff
@@ -98,56 +99,132 @@ def _setup_config_parser_options(self):
...
+                                                        help="Setup user and "
+                                                        "directory for build "
+                                                        "scripts")
+        self._add_setup_environment_args(parser_setup_environment)
+
+        # create the parser for the "build-iso" subcommand
+        parser_iso = subparser.add_parser('build-iso', help="Build ISO image")
+        self._add_iso_args(parser_iso)
+
+    def _add_build_repo_args(self, subparser):
```
farosas: 'subparser' here is inconsistent with the other methods' arguments. I'd prefer if we'd call all of them 'subparser', actually.

host_os.py
```diff
@@ -0,0 +1,50 @@
...
+import logging
+import os
+import sys
+
+from lib import utils
+from tools import build_iso
+from tools import build_package
+from tools import create_release_notes
+from tools import setup_environment
+from tools import upgrade_versions
```
farosas, line 24: I'm fine with `tools`, but just to leave it registered, I think another good option would be `lib/commands`.

lib/iso_spinner.py
```diff
@@ -0,0 +1,138 @@
...
+        self._create_spin_repo()
+
+        self._create_spin_kickstart()
+
+    def _create_spin_repo(self):
+        LOG.info("Creating spin repository inside chroot")
+
+        LOG.debug("Creating spin repo directory")
+        mock_spin_repo_dir = self.config.get('mock_spin_repo').get('dir')
+        self._run_mock_command("--shell 'mkdir %s'" % mock_spin_repo_dir)
```
ghost: Isn't it better to use `mkdir -p` here?

lib/config.py
```diff
@@ -171,6 +160,30 @@ def _add_package_args(self, parser_package):
                                     help='Directory where to clone code repositories',
                                     default='/var/lib/host-os/repositories')
 
+    def _add_release_notes_args(self, parser_relase_notes):
+        """
+        Configures the argument subparser release-notes object to match the
+        expected configuration.
+        """
+        parser_relase_notes.add_argument('--release-notes-repo-url',
```
ghost: Typo: `parser_release_notes`

tools/create_release_notes.py
```diff
@@ -127,12 +125,11 @@ def publish_release_notes(
...
 
-def main(args):
-    CONF = utils.setup_default_config()
+def run(CONF):
     utils.setup_versions_repository(CONF)
 
-    packages_names = (CONF.get('default').get('packages')
-                      or config.discover_packages())
+    packages_names = (CONF.get('default').get('packages') or
+                      config.discover_packages())
```
ghost: PEP8 recommends breaking lines before binary operators: https://www.python.org/dev/peps/pep-0008/#id20

host_os.py
```diff
@@ -19,24 +19,30 @@
...
 if __name__ == '__main__':
     CONF = utils.setup_default_config()
     subcommand = CONF.get('default').get('subcommand')
 
-    if os.getuid() is 0:
+    if os.getuid() is 0 and subcommand != 'set-env':
         print("Please, do not run this script as root, run "
-              "setup_environment.py script in order to properly setup user and"
-              " directory for build scripts")
+              "host_os.py set-env --user <YOUR_USER_LOGIN> script in order to "
```
ghost: Replace "script" with "command" in the two lines above.

lib/config.py
```diff
@@ -136,6 +136,13 @@ def _setup_config_parser_options(self):
         self._add_push_repo_args(parser_upgrade_versions)
         self._add_build_repo_args(parser_upgrade_versions)
 
+        # create the parser for the "set-env" subcommand
+        parser_setup_environment = subparser.add_parser('set-env',
+                                                        help="Setup user and "
+                                                        "directory for build "
+                                                        "scripts")
```
ghost: Breaking the line after `(` to set the `help` parameter in a single line would increase readability.

lib/config.py
```diff
@@ -196,6 +203,16 @@ def _add_push_repo_args(self, subparser):
...
 
+    def _add_setup_environment_args(self, parser_setup_environment):
+        """
+        Configures the argument subparser set-env object to match the expected
+            configuration.
+        """
+        parser_setup_environment.add_argument('--user', '-u',
+                                              help='User login that will run '
+                                              'Host OS commands',
+                                              required=True)
```
ghost: The comment above on line breaking is also valid here.

host_os.py
```diff
@@ -19,24 +19,30 @@
...
+              "properly setup user and directory for build scripts")
+        sys.exit(3)
+
+    if os.getuid() is not 0 and subcommand == 'set-env':
+        print("%s should be run with root privileges" % " ".join(sys.argv))
```
ghost: You could simply print "The set-env command should be run with root privileges"

ghost (changes requested):

lib/config.py
gyribeiro: Yes, moved it to build-package args

lib/config.py
gyribeiro: Done.

lib/config.py
gyribeiro: Not applicable after changes from previous comment.

lib/config.py
gyribeiro: Not applicable after changes from previous comment.

host_os.py
gyribeiro: ok

lib/iso_spinner.py
gyribeiro: Done.

lib/config.py
gyribeiro: Not applicable after last changes.

lib/config.py
gyribeiro: Not applicable after last changes.

lib/config.py
gyribeiro: Not applicable after last changes.

tools/create_release_notes.py
gyribeiro: Done

host_os.py
gyribeiro: Done

host_os.py
gyribeiro: Done

farosas (approved):

lib/config.py
```diff
@@ -24,6 +24,74 @@
...
+    '--build-version':
+        dict(help='Select build version from versions repository'),
+    '--build-versions-repo-dir':
+        dict(help='Directory to clone the build versions repository',
+             default='./components'),
+    '--http-proxy':
+        dict(help='HTTP proxy URL'),
+}
+PACKAGE_ARGS = {
+    '--packages -p':
```
ghost: I'd prefer to use a tuple here. This way, we'll not need to split the string afterwards.
Tuples are immutable and hence can be used as dict keys.
`('--packages', '-p'): dict(...)`

lib/config.py
```diff
@@ -101,59 +170,28 @@ def _setup_config_parser_options(self):
...
+    def _add_subparser(self):
+        subparser = self.parser.add_subparsers(
+            dest="subcommand",
+            help="Available subcommands")
 
+        for command, help_msg, arg_groups in SUBCOMMANDS:
+            parser_command = subparser.add_parser(command, help=help_msg)
+            for arg_group in arg_groups:
+                for arg, options in arg_group.items():
+                    arg = arg.split()
```
ghost: If you follow my tuple suggestion, remove this line.
Also, rename to `args`, since there may be more than one.

lib/config.py
ghost: To make a tuple with a single element, leave a comma at the end: `('--keep-builddir',)`

gyribeiro: Done.

lib/config.py
gyribeiro: Done.

ghost (approved):

lib/iso_spinner.py
```diff
@@ -0,0 +1,146 @@
...
+    def __init__(self, config):
+        self.config = config.get("iso")
+        (self.distro, self.version, self.arch) = distro_utils.detect_distribution()
+        # only the major number must be considered
+        self.version = self.version.split('.')[0]
+
+    def _run_mock_command(self, cmd):
+        try:
+            utils.run_command("mock -r %s %s" % (self.config.get('mock_config'),
+                                                 cmd))
```
mopsfelder: Shouldn't we pass mock_args to this mock command as well?

gyribeiro: Done

gyribeiro:
This closes #50

Issue #112, opened 2016-11-16
Use full commit ids to avoid sha collision/ambiguity


maurorodrigues:
Our CI failled with the follow traceback:
```
kernel: Checking out reference 59349a3
Traceback (most recent call last):
  File "host-os-build.py", line 57, in <module>
    sys.exit(main(sys.argv[1:]))
  File "host-os-build.py", line 43, in main
    bm()
  File "/home/user/jenkins_workspace/workspace/build-host-os/lib/build_manager.py", line 36, in __call__
    packages_class=RPM_Package, distro=self.distro)
  File "/home/user/jenkins_workspace/workspace/build-host-os/lib/packages_manager.py", line 47, in prepare_packages
    package.download_files()
  File "/home/user/jenkins_workspace/workspace/build-host-os/lib/package.py", line 88, in download_files
    self._download_source_code()
  File "/home/user/jenkins_workspace/workspace/build-host-os/lib/package.py", line 101, in _download_source_code
    branch=CONF.get('default').get('branch'))
  File "/home/user/jenkins_workspace/workspace/build-host-os/lib/package.py", line 144, in _setup_repository
    self.repository.checkout(self.commit_id or self.branch or branch)
  File "/home/user/jenkins_workspace/workspace/build-host-os/lib/repository.py", line 92, in checkout
    self.head.reference = self._get_reference(ref_name)
  File "/home/user/jenkins_workspace/workspace/build-host-os/lib/repository.py", line 115, in _get_reference
    return self.commit(ref_name)
  File "/usr/lib/python2.7/site-packages/git/repo/base.py", line 426, in commit
    return self.rev_parse(text_type(rev) + "^0")
  File "/usr/lib/python2.7/site-packages/git/repo/fun.py", line 183, in rev_parse
    obj = name_to_object(repo, rev[:start])
  File "/usr/lib/python2.7/site-packages/git/repo/fun.py", line 89, in name_to_object
    hexsha = short_to_long(repo.odb, name)
  File "/usr/lib/python2.7/site-packages/git/repo/fun.py", line 70, in short_to_long
    return bin_to_hex(odb.partial_to_complete_sha_hex(hexsha))
  File "/usr/lib64/python2.7/site-packages/gitdb/db/base.py", line 265, in partial_to_complete_sha_hex
    raise AmbiguousObjectName(partial_hexsha)
gitdb.exc.AmbiguousObjectName: 59349a3
```

PR #113, opened 2016-11-16
Use full commit SHA-1 when upgrading versions

ghost:
The commit IDs were updated in versions repository to use the full
SHA-1, so this keeps them consistent.


ghost:
This, together with open-power-host-os/versions#57, fixes #112.

maurorodrigues (approved):

PR #115, opened 2016-11-21
Fix scheduler dropping packages when there are dependencies

ghost:
Instead of building all packages, the scheduler could drop some
depending on the order they were listed.

Example:
packages: A, B, C
A depends on B
Scheduler would process first A's dependency B, adding it to the list.
It would then process A and proceed to process B. On processing B, it
would see it was already in the list and would stop, dropping the tail
of the packages list. It would then never process C.

The issue started happening with commit
e79cbbd5fb7dae104b1e4f1adfe62866f7f90c44, because previously the
dependencies were treated differently and could not be passed in the
list of packages.


maurorodrigues:
it works, thanks

PR #116, opened 2016-11-22
Add all open-power-host-os packages to automatic update list

ghost:
All package repositories under the organization should be kept up-to-date.

### DIFF
--- a/lib/package.py
+++ b/lib/package.py
@@ -89,10 +89,10 @@ class Package(object):
         return self.name
 
 
-    def download_files(self):
+    def download_files(self, recurse=True):
         """
         Download package source code and build files.
-        Do the same for its dependencies, recursively.
+        Optionally, do the same for its dependencies, recursively.
         """
         if self.clone_url:
             self._download_source_code()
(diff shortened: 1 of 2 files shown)


ghost:
This depends on open-power-host-os/versions#62

lib/package.py
```diff
@@ -89,19 +89,20 @@ def __repr__(self):
         return self.name
 
 
-    def download_files(self):
+    def download_files(self, recurse=True):
```
maurorodrigues, line 92: I think you meant recursive not recurse (which in pr-br is 'recurso' )

maurorodrigues: other than that +1..

ghost: The verb is form is ["recurse"](https://en.wiktionary.org/wiki/recurse). I can change to "recursive" or "recursively" if you prefer.

maurorodrigues: No need, if you're right you're right. I swear I translated it before being pedantic =).. +1

maurorodrigues (approved):

PR #117, opened 2016-11-23
Add HTTP Proxy support

ghost:
This PR adds the following features:

- `--http-proxy` command line argument
- Sets `http_proxy` and `https_proxy` environment variables for the process (This allows things like `wget` and `curl` to work properly behind a proxy)
- Cloning git repository behing a proxy: This uses a manual git clone command, as GitPython's API does not support a proxy option.

### DIFF
--- a/lib/config.py
+++ b/lib/config.py
@@ -128,6 +128,8 @@ class ConfigParser(object):
         self.parser.add_argument('--log-size',
                                  help='Size in bytes above which the log file '
                                  'should rotate', type=int)
+        self.parser.add_argument('--http-proxy',
+                                 help='HTTP proxy URL')
         self.parser.add_argument('--push-repo-url',
                                  help='URL of the repository used for pushing')
         self.parser.add_argument('--push-repo-branch',
--- a/lib/repository.py
+++ b/lib/repository.py
@@ -20,8 +20,11 @@ import utils
 import git
 import gitdb
 
+from lib import config
 from lib import exception
 
+
+CONF = config.get_config().CONF
 LOG = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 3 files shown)


lib/repository.py
```diff
@@ -43,20 +46,33 @@ def get_git_repository(name, remote_repo_url, parent_dir_path):
...
         LOG.info("Cloning repository from '%s' into '%s'" %
                  (remote_repo_url, repo_path))
         try:
-            return super(GitRepository, cls).clone_from(
-                remote_repo_url, repo_path, *args, **kwargs)
+            if proxy:
+                git_cmd = git.cmd.Git()
+                git_cmd.execute(['git',
+                                 '-c http.proxy=',
+                                 '\'{}\''.format(proxy),
```
ghost: It's a bit more legible if you enclose strings that have single quotes with double quotes: `"'{}'"`.
Also, shouldn't it be separated like this? `'-c', "http.proxy='{}'".format(proxy),`

lib/config.py
```diff
@@ -128,6 +128,9 @@ def _setup_config_parser_options(self):
         self.parser.add_argument('--log-size',
                                  help='Size in bytes above which the log file '
                                  'should rotate', type=int)
+        self.parser.add_argument('--http-proxy',
+                                 help='HTTP proxy URL',
+                                 default=None)
```
ghost: `default=None` can be omitted

host-os-build.py
```diff
@@ -29,6 +29,11 @@
 
 def main(args):
     CONF = utils.setup_default_config()
+    http_proxy = CONF.get('http_proxy')
+
+    if http_proxy:
+        utils.set_http_proxy_env(http_proxy)
+
```
ghost: Shouldn't the proxy be usable for other scripts too? If so, place this inside setup_default_config().

ghost (changes requested):

lib/repository.py
ghost: Done.

lib/config.py
ghost: Removed

host-os-build.py
ghost: Good point. Moved it to setup_default_config

ghost (approved):

PR #119, opened 2016-11-28
Speed up build script by initializing mock chroot only once

ghost:
When the chroot is cleaned at the end of a package build, mock still
keeps this initilized chroot and unpacks it in the next execution.

Also, rename method _prepare_chroot to _copy_files_to_chroot, to clarify
what it does.


ghost:
Unfortunately, my test showed no significant improvement in build time. This might be because YUM keeps a cache and the mock --init executes very fast.

farosas (approved):

PR #120, opened 2016-11-28
Add version and spec release information to release notes

ghost:


lib/rpm_package.py
```diff
@@ -130,3 +130,11 @@ def _load(self):
                     distro_version=self.distro.version)
         except TypeError:
             raise exception.PackageDescriptorError(package=self.name)
+
+    @property
+    def version(self):
+        return self.spec_file.query_tag("version")
```
farosas, line 182: We will start a process and open the file every time we check this property. I think we should cache this value.

lib/rpm_package.py
```diff
@@ -130,3 +130,11 @@ def _load(self):
...
+
+    @property
+    def release(self):
+        return self.spec_file.query_tag("release").split('.')[0]
```
farosas: Same here.

Also, the release field should not be meaningful in the context of an rpm without the whole string. That property was not being used in upgrade_versions.py so either remove it or return the whole Release string.

lib/rpm_package.py
```diff
@@ -15,16 +15,81 @@
...
+    def query_tag(self, tag):
+        return utils.run_command(
+            "rpmspec --srpm -q --qf '%%{%s}' %s 2>/dev/null" % (
+                tag.upper(), self.path)).strip()
+
+    def update_version(self, new_version):
+        LOG.info("Updating '%s' version to: %s" % (self.path, new_version))
+
+        with open(self.path, 'r+') as f:
+            content = f.read()
```
farosas: Opening the file here is not optimal. I'm adding code to update the 'prerelease' tag and this will be duplicated. Perhaps we should have `read_spec` and `write_spec` methods and keep the content of the file in an attribute so that we read it and write it only once. We could also delay the reading until a functions needs it. What do you think?

upgrade_versions.py
```diff
@@ -130,9 +104,10 @@ def update(self, user_name, user_email):
 
         self._read_version_from_repo(pkg.repository.working_tree_dir)
 
-        result = rpm_cmp_versions(self._spec_version, self._repo_version)
+        result = rpm_package.compare_versions(
+            self._spec_version, self._repo_version)
         if result < 0:
-            self._update_version()
+            pkg.spec_file.update_version()
```
farosas: Missing the version here.

upgrade_versions.py
```diff
@@ -186,13 +129,12 @@ def _bump_release(self, pkg, log=None, user_name=None, user_email=None):
         if log:
             assert user_name is not None
             assert user_email is not None
-            rpm_bump_spec(pkg.specfile, log, user_name, user_email)
+            pkg.spec_file.bump_release(log, user_name, user_email)
 
     def _read_spec(self):
```
farosas: I don't think this method makes sense anymore. Just make the calls directly to `self.pkg.version`/`release` or move these attributions to the constructor.

farosas (changes requested):
Good job creating the SpecFile class, it's a nice improvement over the previous code.

lib/rpm_package.py
farosas: fyi: https://github.com/fabianorosas/builds/commits/prerelease

upgrade_versions.py
ghost: Agreed. Changed to `self.pkg.version/release`

lib/rpm_package.py
ghost: I've created a property `content` to handle content manipulation and a method `write_content` to commit to the file.

lib/rpm_package.py
ghost: Agreed. Added a cache to the `query_tag` method.

lib/rpm_package.py
ghost: Ok, returning the whole string since it will be used in create_release_notes.py.

lib/rpm_package.py
```diff
@@ -15,16 +16,103 @@
...
+        return self._content
+
+    @content.setter
+    def content(self, value):
+        """
+        Set the cached content of the spec file.
+        """
+        self._content = value
+
+    def write_content(self):
```
farosas, line 61: Shouldn't this method reset `self._content` as well?

ghost: I don't think so. After executing `write_content`, we are sure that the cache and the file are synchronized. If we reset `self._content` here, it would be re-read the next time we call `self.content` unnecessarily.

farosas: Ok, but we might then introduce a bug when using rpmdev-bumpspec which also writes to the spec file. Calling `bump_release` in between calls to `write_content` would lose the bump information.

ghost: Good catch. I've forced the bump_release method to set `self._content = None`, so it gets reloaded when needed.

lib/rpm_package.py
```diff
@@ -15,16 +16,117 @@
...
+    def update_prerelease_tag(self, new_prerelease):
+        self.content = self._replace_macro_definition(
+                'prerelease', new_prerelease, self.content)
+        self.write_content()
+        LOG.info("Updated '%s' prerelease tag to: %s"
+                 % (self.path, new_prerelease))
+
+    def _replace_macro_definition(self, macro_name, replacement, text):
+        return re.sub(r'(%%define\s+%s\s+)\S+' % macro_name,
+                      r'\g<1>' + replacement, text)
```
farosas: We could remove the `text` argument and use `self.content` directly here.

farosas (approved):

PR #121, opened 2016-11-29
Support -rc tags in upgrade_versions.py

farosas:


upgrade_versions.py
```diff
@@ -214,8 +234,17 @@ def _read_version_from_repo(self, repo_path):
 
                 with open(version_file, 'r') as f:
                     match = re.search(regex, f.read())
+                    match_groups = list(match.groups())
```
ghost: This line doesn't need to be in the `with` block.

upgrade_versions.py
```diff
@@ -188,6 +193,21 @@ def _bump_release(self, pkg, log=None, user_name=None, user_email=None):
...
+        if not self._prerelease:
+            self._prerelease = "%{nil}"
+
+        try:
+            with open(self.pkg.specfile, 'r+') as f:
+                content = self._replace_macro_definition(
+                    'prerelease', self._prerelease, f.read())
+                f.seek(0)
+                f.write(content)
+                LOG.info("%s: Updated prerelease tag" % self.pkg)
```
ghost: This line doesn't need to be in the `with` block.

upgrade_versions.py
```diff
@@ -188,6 +193,21 @@ def _bump_release(self, pkg, log=None, user_name=None, user_email=None):
             assert user_email is not None
             rpm_bump_spec(pkg.specfile, log, user_name, user_email)
 
+    def _update_prerelease_tag(self):
+        if not self._prerelease:
+            self._prerelease = "%{nil}"
```
ghost: Why do we need to set to `"%{nil}" here?` If we don't have a pre-release, no string will be replaced in the file. You could initialize `self._prerelease` to an empty string just to avoid exceptions below.

upgrade_versions.py
```diff
@@ -188,6 +193,21 @@ def _bump_release(self, pkg, log=None, user_name=None, user_email=None):
...
+
+        try:
```
ghost: We don't really need the try block. `re.sub` does not raise exceptions if no strings are found, it just returns the unaltered string.

upgrade_versions.py
```diff
@@ -214,8 +234,17 @@ def _read_version_from_repo(self, repo_path):
 
                 with open(version_file, 'r') as f:
                     match = re.search(regex, f.read())
+                    match_groups = list(match.groups())
+
+                last_group = match_groups[-1]
+                for term in PRERELEASE_TERMS:
+                    if term in last_group:
+                        self._prerelease = last_group.replace('-','.')
```
ghost: Why do you need this replace?

ghost (changes requested):

upgrade_versions.py
farosas: For when the version changes from '4.9.0-rc4' to '4.9.0'. Not replacing would leave the '.rc4' there.

upgrade_versions.py
farosas: The prerelease in kernel's Makefile is '-rc4'. We can't have the '-' in the Release field of the spec file.

upgrade_versions.py
ghost: Got it. I think you could initilize to `%{nil}` in the constructor then.

maurorodrigues (approved):

ghost (approved):

PR #123, opened 2016-11-29
Enable tmpfs plugin in mock

mopsfelder:
Add `--mock-args` switch to `host-os-build.py` script as well as its test cases.  And fix a minor issue with a missing module in `requirements-dev.txt`.

* 41f57ae Add test cases for --mock-args parameter
* a3643d8 Add PyYAML in requirements-dev.txt
* 04f23f1 Pass mock_args to mock command
* 7159e91 Add mock_args option in config.yaml

### DIFF
--- a/tests/unit/config_tests.py
+++ b/tests/unit/config_tests.py
@@ -20,6 +20,7 @@ class TestConfigParser(unittest.TestCase):
         ('--keep-builddir', 'keep_builddir', True),
         ('--build-versions-repository-url=foo', 'build_versions_repository_url', 'foo'),
         ('--build-version=foo', 'build_version', 'foo'),
+        ('--mock-args=foo', 'mock_args', 'foo'),
     ])
     def test_parse_arguments_list_WithLongArgument_ShouldParseArgumentValue(self, argument, key, expected):
         cfg = ConfigParser()
@@ -39,6 +40,7 @@ class TestConfigParser(unittest.TestCase):
         ('keep_builddir', False),
         ('build_versions_repository_url', None),
         ('build_version', None),
+        ('mock_args', ''),
     ])
     def test_parse_arguments_list_WithoutArgument_ShouldUseDefaultValue(self, key, expected):
         cfg = ConfigParser()
--- a/config.yaml
+++ b/config.yaml
@@ -9,3 +9,8 @@ default:
  arch_and_endianness: 'ppc64le'
  release_notes_repo_url: "https://github.com/open-power-host-os/open-power-host-os.github.io.git"
  release_notes_repo_branch: "master"
+ mock_args:
+   "--enable-plugin=tmpfs
+   --plugin-option=tmpfs:keep_mounted=True
+   --plugin-option=tmpfs:max_fs_size=50g
+   --plugin-option=tmpfs:required_ram_mb=4096"
--- a/lib/config.py
+++ b/lib/config.py
@@ -118,6 +118,9 @@ class ConfigParser(object):
         self.parser.add_argument('--keep-builddir',
                                  help='Keep build directory and its logs and '
                                  'artifacts.', action='store_true')
+        self.parser.add_argument('--mock-args',
+                                 help='Arguments passed to mock command',
+                                 default='')
         self.parser.add_argument('--build-versions-repository-url',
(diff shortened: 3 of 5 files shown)


ghost:
Looks good to me.

It would have been nice to have some tests, though.

farosas (approved):

ghost (approved):

tests/unit/config_tests.py
```diff
@@ -20,6 +20,8 @@ class TestConfigParser(unittest.TestCase):
         ('--keep-builddir', 'keep_builddir', True),
         ('--build-versions-repository-url=foo', 'build_versions_repository_url', 'foo'),
         ('--build-version=foo', 'build_version', 'foo'),
+        ('--mock-args=--enable-plugin=tmpfs --plugin-option=tmpfs:keep_mounted=True', 'mock_args',
```
farosas: I think it's a bit misleading to use the options in this way inside a single string. The test will pass and there is nothing to do with the fact that there is two mock arguments there. AFAIK, this is not how argparse will return the values.

farosas: @snowcrshd does this make sense to you?

ghost: @fabianorosas not sure I get what you mean. Are you saying that testing with explicit mock arguments might cause someone to think those arguments will be parse as well?

If so, I agree with you.

For example:

Is `--plugin-option=tmpfs:keep_mounted=True` a parameter given to mock? Or is it a parameter parsed by the host os config parser? It is a bit confusing. Shouldn't the mock-args parameters be surrounded by quotes, since they are separated by spaces?

farosas: @snowcrshd, 
> Shouldn't the mock-args parameters be surrounded by quotes, since they are separated by spaces?

I refer to this ^

This test doesn't check that it works with more than one "mock-arg" although it looks like it does.

But, yeah, I would also change the names to '--foo' '--bar'

ghost: The problem with passing a parameter like `'--plugin-option=tmpfs:keep_mounted=True'` is that argparse will think its a parameter it should be parsing and will raise an exception. By bundling the parameters in a single string, it correctly identifies it as a parameter value. I think another option that works would be adding a space before the parameter: `' --plugin-option=tmpfs:keep_mounted=True'`

farosas: @olavphibm Wow that is some weird behavior there. We shouldn't be using 'nargs', then. Let's just take a string. From what I gathered, our usage of argparse is returning `['--enable-plugin=... --plugin-option=...']` and we later do `" ".join(mock_args)`, which works, but is misleading.

(We should still use '--foo --bar' for the tests instead of the real options, though.)

ghost: Indeed, that might be better to avoid confusion. If so, we should change the yaml to a string too.

mopsfelder: Let me explain my rationale here.

I have thought in `mock_args` as a list to keep `config.yaml` more readable so humans could clearly see parameters passed to `mock_args`, one per line:

```
mock_args:
   - --enable-plugin=tmpfs
   - --plugin-option=tmpfs:keep_mounted=True
   - --plugin-option=tmpfs:max_fs_size=50g
   - --plugin-option=tmpfs:required_ram_mb=4096
```

Otherwise, it would have to be a very long line in `config.yaml`, damaging readability:

```mock_args: "--enable-plugin=tmpfs --plugin-option=tmpfs:keep_mounted=True --plugin-option=tmpfs:max_fs_size=50g --plugin-option=tmpfs:required_ram_mb=4096"```

From what I read, you seem to prefer `mock_args` as a simple string instead of a list.  I will change that.

I will also use metasyntactic names in the unit tests.

I did appreciate your input.  Thank you!

farosas: @mopsfelder Yes, you choice makes perfect sense, but I think we will have no option here because of the way argparse works (and to keep cmdline and config file inputs consistent). To avoid losing readability perhaps you would like to play with the various ways of specifying strings in YAML: http://stackoverflow.com/a/21699210

mopsfelder: Thanks, @fabianorosas!

ghost (approved):

PR #129, opened 2016-12-19
Change argument dropping to allow missing args in config

ghost:
This was noticed when removing the `mock_args` option entirely
from the config file.

The `None` value would not be dropped and it would break the builds
scripts.

### DIFF
--- a/lib/config.py
+++ b/lib/config.py
@@ -186,7 +186,7 @@ class ConfigParser(object):
 
         # drop None values
         for key, value in args.items():
-            if not value:
+            if value is None:
                 args.pop(key)
 
         config['default'].update(args)


ghost:
I'd rephrase the commit message to make it clear that we are allowing _empty_ args (`""`), not _missing_ (`None`) args:

Change argument dropping to allow empty args in config

This was noticed when removing the `mock_args` option entirely from the config file.

The empty string (default value for `mock_args`) would be dropped and it would break the builds scripts.

ghost (approved):

ghost:
start build

snowcrshd:
start build

snowcrshd:
test plz

PR #131, opened 2016-12-21
Update mock's tmpfs plugin parameters with realistic values

ghost:
We had a misconception on the meaning of the parameters to the tmpfs
plugin. The "required_ram_mb" parameter must be greater than the
"max_fs_size" parameter, to make sure the system has enough RAM to mount
the whole tmpfs. Otherwise, the build may try to use more memory than
what is available and crash with an "out of memory" failure.

We've found out that guests running the whole build grow their
expandable virtual disks to aproximately 30GB, so the max file system
size must be at least that. With a little slack and adding 8GiB for
"normal" RAM usage, we get to the defined value of approximately 40GiB
of required RAM to enable the tmpfs plugin.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -10,10 +10,11 @@ default:
  release_notes_repo_url: "https://github.com/open-power-host-os/open-power-host-os.github.io.git"
  release_notes_repo_branch: "master"
  mock_args:
+   # If the system has less RAM than required, the plugin will be disabled
    "--enable-plugin=tmpfs
    --plugin-option=tmpfs:keep_mounted=True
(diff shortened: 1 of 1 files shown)


farosas (approved):

ghost:
Looks good.

gyribeiro (approved):

alanoe (approved):
I'd comment somewhere in the config.yaml that mock will work fine if there is less RAM available than set in required_ram_mb, but then it will not use tmpfs. It is not obvious

lisibann (approved):

Issue #38, opened 2016-08-31
Make yaml descriptors cleaner


maurorodrigues:
There are some elements that shouldn't be necessary to specify in the yaml assuming we respect some standards, like distro specific values: 
- rpmmacro file should stay in the same place always
- same for spec
- same for sources

So let's refactor some code to auto discover whatever is possible.

PR #132, opened 2016-12-21
Make expects_source attribute optional in package yaml file

alanoe:


ghost (approved):

ghost:
start tests

gyribeiro (approved):

alanoe:
start tests

ghost:
start tests

lib/package.py
```diff
@@ -161,6 +161,9 @@ def _download_source(self, build_dir):
         An alternative to just execute a given command to obtain sources.
         """
         utils.run_command(self.download_source, cwd=build_dir)
+        # automatically append tar.gz if expects_source has no extension
+        if self.expects_source == self.name:
+            self.expects_source = "%s.tar.gz" % self.name
```
ghost, line 166: Appending the extension was never needed here, why do we need it now?
I see the `BuildManager` calls `prepare_packages()`, which downloads the source code.
The `BuildManager` then proceeds to build, which reaches the `Mock` class `build()`, `_prepare()` and `_prepare_archive` methods, and finally `GitRepository.archive()`. This method then uses the `expects_source`, which was passed as the parameter `archive_name` as a prefix to the tar.gz file, and appends the extension. It appears to me that this new code would make the file have the extension duplicated.

ghost (changes requested):

lib/package.py
alanoe: GitRepository.archive() and Package._download_source() are mutually exclusive. This patch only alters _download_source(), which uses commands in download_source attribute in package yaml to download package source code. 

The only packages which use this attribute are gcc, libguestfs and kubernetes. This patch was not needed before because the expects_source attribute hardcoded in the packages YAML contained the tar.* extension. The expects_source attribute is now set to be the package name at Package constructor by default and therefore the extension must be added at some point when the default value is used.

An alternative would be to add the tar.gz extension at the constructor, but then other part of the code which clones from a repository instead of allowing custom download commands would need to be updated.

ghost (approved):

farosas:
This relates to #38

PR #133, opened 2016-12-22
Use LBYL instead of EAFP in directory creation in setup_environment.py

alanoe:
Newer code is simpler. The only disadvantage is that it
is not thread-safe, but this code will never be run concurrently,
so this is not an issue.

### DIFF
--- a/tools/setup_environment.py
+++ b/tools/setup_environment.py
@@ -20,7 +20,6 @@ from lib import exception
 from lib import utils
 
 USER_EXISTS = 9
-DIRECTORY_EXISTS = 17
 
 
 def setup_user(user):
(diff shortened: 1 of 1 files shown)


ghost (approved):

gyribeiro (approved):

ghost:
start tests

tools/setup_environment.py
```diff
@@ -44,11 +43,8 @@ def setup_user(user):
 
 
 def setup_directory(directory, uid, gid):
-    try:
+    if not os.path.exists(directory):
```
ghost: Use `os.path.isdir` here, `exists` would allow a file with this name.

ghost (changes requested):

ghost:
start tests

ghost (approved):

PR #134, opened 2016-12-23
Make archive sub attrib optional in packages sources attributes in package YAML file

alanoe:


lib/repository.py
```diff
@@ -41,6 +42,13 @@ def get_git_repository(name, remote_repo_url, parent_dir_path):
     dir, according to specified name.
     If it does not exist, clone it from the specified URL.
     """
+    # infer name if it is not provided
+    if not name:
+        match_obj = re.match(".*/([a-zA-Z0-9\-]*)\.git", remote_repo_url)
```
ghost: Won't this break if the repository name is some weird string?

There are some [special characters that are allowed in a URL](http://stackoverflow.com/questions/1856785/characters-allowed-in-a-url#1856809) that could show up in the repository's name.

ghost (changes requested):
I'd rather not pass in 'None' into all those function calls, just to have an 'if' checking if the name was provided.

Better to just remove the 'name' parameter and either infer it or fail noisily.

lib/repository.py
ghost: I'd use `urlparse` combined with `os.path.basename` for that.

olavphibm:
start build

ghost:
I'm with either removing the "name" parameter completely, or leaving it as a keyword arg, so we don't have to pass `None` every time.

olavphibm:
start build

lib/repository.py
alanoe: imho [a-zA-Z0-9\-_]* is a safe regex, I have never seen other chars in source code repositories names

alanoe:
OH> I'm with either removing the "name" parameter completely, or leaving it as a keyword arg, so we don't have to pass None every time.

Agreed. I see no reason to have this argument in our build code, so I'll remove it.

lib/repository.py
```diff
@@ -35,12 +36,15 @@ def __init__(self, push_info):
...
+def get_git_repository(remote_repo_url, parent_dir_path):
     """
-    Get a local git repository located in a subdirectory of the parent
-    dir, according to specified name.
+    Get a local git repository located in a subdirectory of the parent directory.
     If it does not exist, clone it from the specified URL.
     """
+    # infer name
+    url_parts = urlparse.urlparse(remote_repo_url)
+    name = os.path.basename(url_parts.path)
```
ghost: This name might have a ".git" that we need to remove.

lib/repository.py
```diff
@@ -35,12 +36,15 @@ def __init__(self, push_info):
         super(PushError, self).__init__(message)
 
 
-def get_git_repository(name, remote_repo_url, parent_dir_path):
+def get_git_repository(remote_repo_url, parent_dir_path):
     """
-    Get a local git repository located in a subdirectory of the parent
-    dir, according to specified name.
+    Get a local git repository located in a subdirectory of the parent directory.
```
ghost: You could mention it is "named after the base name of the URL path"

lib/versions_repository.py
```diff
@@ -15,12 +15,12 @@ def setup_versions_repository(config):
     Clone and checkout the versions repository and halt execution if
     anything fails.
     """
-    path, dir_name = os.path.split(
+    path, _ = os.path.split(
```
ghost, line 18: `os.path.dirname` could be used here instead of `split`

ghost (changes requested):

lib/versions_repository.py
alanoe: I prefer to not alter how parent directory path is obtained in this patch. This is an improvement which could be done in a separate patch.

lib/repository.py
alanoe: Done

lib/repository.py
alanoe: Done

lib/repository.py
ghost: I think that `# infer name` could be removed.

It doesn't add much information to that context. It ends up being just noise in the source code.

fabianorosas:
start build

alanoe:
start build

lib/repository.py
```diff
@@ -35,12 +36,16 @@ def __init__(self, push_info):
...
-def get_git_repository(name, remote_repo_url, parent_dir_path):
+def get_git_repository(remote_repo_url, parent_dir_path):
     """
-    Get a local git repository located in a subdirectory of the parent
-    dir, according to specified name.
+    Get a local git repository located in a subdirectory of the parent directory,
+    named after the file name of the URL path (git default).
     If it does not exist, clone it from the specified URL.
     """
+    # infer name
```
ghost: Please remove this comment, it doesn't add meaningful information.

ghost:
start build

ghost (approved):

ghost:
start build

farosas:
This relates to #38

lib/repository.py
alanoe: IMHO considering a comment to be obvious (and thus useless) or not is usually a matter of personal taste. Very minor issue IMHO. Anyway, I have updated the comment to make it more useful.

alanoe:
Pls let us merge this ASAP so I do not need to rebase this patch again.

ghost (approved):

PR #140, opened 2017-01-06
Remove libguestfs from the list of packages added to iso

farosas:
We are no longer building this package, so there is no need to mention
it explicitly. It will be taken from the CentOS extras repository.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -31,7 +31,6 @@ iso:
             - "iprutils"
             - "kernel"
             - "kimchi"
-            - "libguestfs"
             - "libseccomp"
             - "libvirt"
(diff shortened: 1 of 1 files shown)


farosas:
Depends on: open-power-host-os/versions#87
Closes #136

ghost (approved):

ghost:
This actually does not depend on open-power-host-os/versions#87, only the reverse is true, so this could be merged already.

farosas:
> This actually does not depend on open-power-host-os/versions#87, only the reverse is true, so this could be merged already.

Well, if we end up not merging the other one, then we will have to revert this.

PR #141, opened 2017-01-06
Fix cleaning of mock chroot

farosas:
When using the tmpfs plugin, mock creates an extra mount point:

mock_chroot_tmpfs on /var/lib/mock/epel-7-ppc64le/root type tmpfs

This mount point is only unmounted if --clean is run with the tmpfs
plugin enabled.


mopsfelder (approved):

mopsfelder:
That issue was annoying.  Thanks for chasing it down and fixing it!

ghost (approved):

PR #144, opened 2017-01-11
Check if package dependencies are installed in host_os.py

alanoe:
Validate RPM packages dependencies before doing anything else in host_os.py

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -29,22 +29,11 @@ RPM Based distributions
 $ sudo yum install --downloadonly --downloaddir=. epel-release
 $ sudo yum localinstall epel-release-7-5.noarch.rpm # Note the version may change.
 
-* Install
-
- - bzip2
- - git
- - GitPython
- - lzop
- - mock
- - nosync
- - python-lxml
- - PyYAML
- - subversion
- - wget
+* Install RPM dependencies
 
 ::
 
-$ sudo yum install -y bzip2 git GitPython lzop mock nosync python-lxml PyYAML subversion wget
+$ sudo yum install -y $(cat rpm_requirements.txt)
 
 Settings
 --------
--- a/README.rst
+++ b/README.rst
@@ -34,17 +34,17 @@ $ sudo yum localinstall epel-release-7-5.noarch.rpm # Note the version may chang
  - bzip2
  - git
  - GitPython
+ - lzop
  - mock
+ - nosync
+ - python-lxml
  - PyYAML
  - svn
  - wget
- - python-lxml
- - nosync
- - lzop
 
 ::
 
-$ sudo yum install -y bzip2 git GitPython mock PyYAML svn wget
+$ sudo yum install -y bzip2 git GitPython lzop mock nosync python-lxml PyYAML svn wget
 
 Settings
 --------
--- a/README.rst
+++ b/README.rst
@@ -39,12 +39,12 @@ $ sudo yum localinstall epel-release-7-5.noarch.rpm # Note the version may chang
  - nosync
  - python-lxml
  - PyYAML
- - svn
+ - subversion
  - wget
 
 ::
 
-$ sudo yum install -y bzip2 git GitPython lzop mock nosync python-lxml PyYAML svn wget
+$ sudo yum install -y bzip2 git GitPython lzop mock nosync python-lxml PyYAML subversion wget
 
 Settings
 --------
--- a/host_os.py
+++ b/host_os.py
@@ -17,6 +17,8 @@ import os
 import sys
 
 from lib import config
+from lib.exception import SubprocessError
+from lib.utils import run_command
 from tools import build_iso
 from tools import build_package
 from tools import create_release_notes
@@ -31,11 +33,42 @@ SUBCOMMANDS = {
     'set-env': setup_environment,
     'build-iso': build_iso,
 }
+REQUIRED_PACKAGES_FILE_PATH = "rpm_requirements.txt"
+MISSING_PACKAGES_ERROR = 4
+
+def is_package_installed(package_name):
+    """
+    Checks if a RPM package is installed
+    """
+
+    cmd = "rpm -q %s" % package_name
+    try:
+        run_command(cmd, shell=True)
+    except SubprocessError as e:
+        # rpm returns 1 when search string is not found and other non-zero values
+        # if an error occurred
+        #pylint: disable=no-member
+        if e.returncode == 1:
+            return False
+        else:
+            raise
+
+    return True
 
 
 if __name__ == '__main__':
     CONF = config.setup_default_config()
     subcommand = CONF.get('default').get('subcommand')
+
+    # validate if all required packages are installed
+    with open(REQUIRED_PACKAGES_FILE_PATH) as f:
+        required_packages = f.read().splitlines()
+    missing_packages = [p for p in required_packages
+        if not is_package_installed(p)]
+    if missing_packages:
+        print("Following packages should be installed before running this "
+              "script: %s" % ", ".join(missing_packages))
+        sys.exit(MISSING_PACKAGES_ERROR)
 
     if os.getuid() is 0 and subcommand != 'set-env':
         print("Please, do not run this command as root, run "
--- a/rpm_requirements.txt
+++ b/rpm_requirements.txt
@@ -1,0 +1,10 @@
+bzip2
+git
+GitPython
+lzop
+mock
+nosync
+python-lxml
+PyYAML
+subversion
+wget


README.rst
```diff
@@ -34,17 +34,17 @@ $ sudo yum localinstall epel-release-7-5.noarch.rpm # Note the version may chang
...
  - mock
+ - nosync
+ - python-lxml
  - PyYAML
- - svn
+ - subversion
  - wget
- - python-lxml
- - nosync
- - lzop
```
farosas, line 43: I vote for dropping this list. Having two places to update with packages is already too much. What about we keep a file with the dependencies list and suggest a `yum install $(cat dependencies)` instead? Then your code could read from the same file to check if the packages are installed.

host_os.py
```diff
@@ -31,12 +35,41 @@
...
+REQUIRED_PACKAGE_DEPENDENCIES = [
+    "bzip2", "git", "GitPython", "lzop", "mock", "nosync", "python-lxml", "PyYAML", "subversion", "wget"
+]
+
+def is_package_installed(package_name):
+    """
+    Checks if a RPM package is installed
+    """
+
+    cmd = "rpm -qa | grep %s" % package_name
```
farosas: I think `"rpm -q %s" % package_name` ought to suffice. Then we could cleanup the logic below and just check the return code.

farosas (changes requested):

README.rst
alanoe: Keeping a single list is good. +1

host_os.py
alanoe: +1

alanoe:
This Pylint error seems a false positive to me. returncode is defined when SubprocessError is raised in run_command() in lib/utils.py

host_os.py
```diff
@@ -32,11 +34,36 @@
...
+def is_package_installed(package_name):
+    """
+    Checks if a RPM package is installed
+    """
+
+    cmd = "rpm -q %s" % package_name
+    try:
+        run_command(cmd, shell=True)
+    except SubprocessError as e:
+        # grep returns 1 when search string is not found and 2 if an error occurred
```
mopsfelder: grep?

host_os.py
```diff
@@ -32,11 +34,36 @@
...
     subcommand = CONF.get('default').get('subcommand')
 
+    PACKAGE_DEPENDENCIES_FILE_PATH = "rpm_requirements.txt"
+    package_dependencies = open(PACKAGE_DEPENDENCIES_FILE_PATH).readlines()
+    package_dependencies = [x.strip('\n') for x in package_dependencies]
+    for package in package_dependencies:
+        if not is_package_installed(package):
+            print("%s package should be installed before running this script" % package)
+            sys.exit(4)
+
```
mopsfelder, line 72: The `PACKAGE_DEPENDENCIES_FILE_PATH` is a huge variable name and it is used only once.  I'd use a variable if its value comes from a configuration file.  Since its value is hard-coded here, I don't see the point for a variable.

If the system does not have all of the rpm requirements, `host_os.py` will exit with the same error every time user tries to run it until all packages are installed.  I guess it would be better for user to know all package names that there are missing on his system in one shot.

What about defining an error code when packages are not installed, similar to what has been proposed in pull 143?

I'd also use list-comprehension to simplify this block.  For example:

```python
with open("reqs.txt") as f:
    pkg_reqs = [l.strip('\n') for l in f.readlines()]

not_installed = [p for p in pkg_reqs if not is_package_installed(p)]

if not_installed:
    print("The following packages should be installed before running this "
          "script:\n%s" % " ".join(not_installed))
    sys.exit(DEPS_NOT_INSTALLED_ERROR)
```

mopsfelder:
Pull request name should be something easier to understand.  What about `Check if package dependencies are installed in host_os.py`?

host_os.py
ghost: I agree with @mopsfelder, just adding that `readlines` does not return line separators, so the `.strip('\n')` could be removed.

ghost (changes requested):

ghost:
About the Pylint error, the `returncode` is passed as a keyword arg to the `BaseException` superclass and the attribute is created dynamically, so Pylint has no way to know it exists. I suggest you use `# pylint: disable=no-member`.

host_os.py
alanoe: MO> The PACKAGE_DEPENDENCIES_FILE_PATH is a huge variable name and it is used only once. I'd use a variable if its value comes from a configuration file. Since its value is hard-coded here, I don't see the point for a variable.

This is a constant. Although Python does not have the concept of constant, I always prefer to use them when hardcoding strings (eg files names/paths). It is usually a good programming practice and make code reading easier (imho). If other devs agree with you, I may remove the constant.

MO>  I guess it would be better for user to know all package names that there are missing on his system in one shot.

I agree. Done.

MO> What about defining an error code when packages are not installed, similar to what has been proposed in pull 143?

I'm already returning a specific error code (4) for this situation. I'll create a constant for it.

MO> I'd also use list-comprehension to simplify this block. 

This is a matter of taste. I personally like more the traditional for loop, but I know most developers of our team prefer the list comprehension, so I'll use it.

OH> just adding that readlines does not return line separators, so the .strip('\n') could be removed.

I don't think so.

>>> open("bla.txt").readlines()
['a\n', 'b\n', 'c\n', 'd\n']

ghost: I'm sorry, I misread the code. My suggestion is to replace:
`pkg_reqs = [l.strip('\n') for l in f.readlines()]`
with
`pkg_reqs = f.read().splitlines()`

host_os.py
```diff
@@ -31,12 +33,40 @@
...
     CONF = config.setup_default_config()
     subcommand = CONF.get('default').get('subcommand')
 
+    package_dependencies = open(PACKAGE_DEPENDENCIES_FILE_PATH).readlines()
+    package_dependencies = [p.strip('\n') for p in package_dependencies]
+    missing_packages = [p for p in package_dependencies if not is_package_installed(p)]
+    if missing_packages:
+        print("Following packages should be installed before running this script: %s" % ", ".join(missing_packages))
+        sys.exit(MISSING_PACKAGE_DEPENDENCIES_ERROR)
+
```
mopsfelder, line 72: What about fitting that code in 79 columns, as per [PEP 8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length)?

```python
pkg_deps = open(PKG_DEPS_FILE).readlines()
pkg_deps = [p.strip('\n') for p in pkg_deps]
missing_pkgs = [p for p in pkg_deps if not is_package_installed(p)]
if missing_pkgs:
    print("Following packages should be installed before running "
          "this script: %s" % ", ".join(missing_pkgs))
    sys.exit(MISSING_PKG_DEPS_ERROR)
```

With shorter variable name you can still have meaningful names with enhanced readbility.

mopsfelder (changes requested):

host_os.py
```diff
@@ -31,12 +33,40 @@
...
+    return True
 
 
 if __name__ == '__main__':
     CONF = config.setup_default_config()
     subcommand = CONF.get('default').get('subcommand')
 
+    package_dependencies = open(PACKAGE_DEPENDENCIES_FILE_PATH).readlines()
+    package_dependencies = [p.strip('\n') for p in package_dependencies]
```
ghost: Minor: This could be written like this: `pkg_reqs = f.read().splitlines()`

host_os.py
```diff
@@ -31,12 +33,40 @@
...
+            raise
+
+    return True
 
 
 if __name__ == '__main__':
     CONF = config.setup_default_config()
     subcommand = CONF.get('default').get('subcommand')
 
+    package_dependencies = open(PACKAGE_DEPENDENCIES_FILE_PATH).readlines()
```
ghost: You're not closing the file, better use `with`

ghost (changes requested):

alanoe:
start build

host_os.py
alanoe: Done

host_os.py
alanoe: Done

host_os.py
alanoe: IMHO using full words are clear than abbreviations (eg "package" is clearer than "pkg") in variables names. It is a matter of taste.

Code now fits in 79 columns

farosas (approved):

ghost (approved):

rpm_requirements.txt
```diff
@@ -0,0 +1,10 @@
+bzip2
+git
+GitPython
+lzop
+mock
+nosync
+python-lxml
+PyYAML
+subversion
```
mopsfelder, line 9: Has `svn` package changed its name to `subversion`?  Which one is correct?  Or are both correct?

alanoe: The correct rpm package name is subversion. There is a separate commit in the PR which fixes that.

PR #145, opened 2017-01-11
Infer RPM spec, macros and build files attributes automatically

alanoe:


mopsfelder:
Please provide a meaningful pull request name and a short description of what is being proposed.

lib/rpm_package.py
```diff
@@ -140,13 +140,17 @@ def _load(self):
...
-            self.build_files = files.get('build_files', None)
-            if self.build_files:
-                self.build_files = os.path.join(
-                    self.package_dir, self.build_files)
+            files = self.package_data.get('files', {}).get(
+                self.distro.lsb_name, {}).get(self.distro.version, {})
+
+            default_build_files_dir_rel_path = os.path.join(
+                self.distro.lsb_name, self.distro.version, "SOURCES")
+            build_files_dir_rel_path = files.get('build_files', default_build_files_dir_rel_path)
```
ghost: SLOF has an empty YAML attribute `build_files`, which means its value is set to `None`.
To avoid breaking things and be backwards compatible, we need to do something like:
`build_files_dir_rel_path = files.get('build_files') or default_build_files_dir_rel_path`
I know, it sucks.

ghost (changes requested):

lib/rpm_package.py

olavphibm:
start tests

ghost (approved):

lib/rpm_package.py
```diff
@@ -140,13 +140,17 @@ def _load(self):
...
             # load distro files
-            files = self.package_data.get('files').get(
-                self.distro.lsb_name).get(self.distro.version)
-
-            self.build_files = files.get('build_files', None)
-            if self.build_files:
-                self.build_files = os.path.join(
-                    self.package_dir, self.build_files)
+            files = self.package_data.get('files', {}).get(
+                self.distro.lsb_name, {}).get(self.distro.version, {})
```
ghost: The yaml files have `centos` as a subnode, so the self.distro.lsb_name must be converted to lowercase here.

ghost (changes requested):

lib/rpm_package.py
alanoe: It does not make sense to have different cases in different places, centOS should be written with the same case in the entire code to make things simpler. Therefore, it is better if I either use 'centos' in lib/distro_utils.py or use 'centOS' in all yaml files

ghost: The script is not working like this. I'd actually prefer to have "CentOS", but that requires changing the versions repo and we also have to keep compatibility...

mopsfelder: What if directory `centOS` is renamed to `centos` or `CentOS` after Host OS 2.0?

alanoe: We agreed on naming the distro centOS everywhere. The Python class remains to be CentOS, but that is not a problem because there is a dict between distro names and corresponding Python classes.

I am sending patches to handle this. They will keep backwards compatibility, which I hope we can remove on OP Host OS 2.0.

lib/rpm_package.py
```diff
@@ -169,8 +169,11 @@ def _load(self):
             if self.rpmmacro:
                 self.rpmmacro = os.path.join(self.package_dir, self.rpmmacro)
 
-            self.spec_file = SpecFile(
-                os.path.join(self.package_dir, files.get('spec')))
+            default_spec_file_rel_path = os.path.join(
+                self.distro.lsb_name, self.distro.version, "%s.spec" % self.name)
```
ghost, line 182: I believe `distro_attrib_name` should be used here.

lib/rpm_package.py
```diff
@@ -150,10 +150,14 @@ def _load(self):
             files = self.package_data.get('files', {}).get(
                 distro_attrib_name, {}).get(self.distro.version, {})
 
-            self.build_files = files.get('build_files', None)
-            if self.build_files:
-                self.build_files = os.path.join(
-                    self.package_dir, self.build_files)
+            default_build_files_dir_rel_path = os.path.join(
+                self.distro.lsb_name, self.distro.version, "SOURCES")
```
ghost, line 154: I believe `distro_attrib_name` should be used here.

lib/rpm_package.py
```diff
@@ -169,9 +169,14 @@ def _load(self):
                 dep = RPM_Package.get_instance(dep_name, self.distro)
                 self.build_dependencies.append(dep)
 
-            self.rpmmacro = files.get('rpmmacro', None)
-            if self.rpmmacro:
-                self.rpmmacro = os.path.join(self.package_dir, self.rpmmacro)
+            default_rpm_macros_file_rel_path = os.path.join(
+                self.distro.lsb_name, self.distro.version, "rpmmacro")
```
ghost, line 173: I believe `distro_attrib_name` should be used here.

ghost (changes requested):
Mention the name attribute in the PR name/description too.

ghost (approved):
Please disregard my previous comments. I've just realized the name of the versions directories was already `centOS`.

PR #148, opened 2017-01-12
Add options to git source attrib in package yaml file which allows gcc to be downloaded from its mirror git repository

alanoe:


lib/package.py
```diff
@@ -136,9 +136,11 @@ def _load(self):
         self.name = self.package_data.get('name')
         self.sources = self.package_data.get('sources', [])
         for source in self.sources:
+            source_name = source.keys()[0]
             if 'archive' not in source.values()[0]:
-                source_name = source.keys()[0]
                 source[source_name]['archive'] = self.name
+            if 'ref_to_fetch' not in source.values()[0] and source_name == 'git':
+                source[source_name]['ref_to_fetch'] = None
```
ghost: I think its a little more readable to write:
```
if source_name == 'git':
    source[source_name].setdefault('ref_to_fetch', None)
```
The effect is the same as your code.
The same could be applied to the "archive" above, removing the `if`.

lib/package.py
```diff
@@ -139,8 +139,11 @@ def _load(self):
             source_name = source.keys()[0]
             if 'archive' not in source.values()[0]:
                 source[source_name]['archive'] = self.name
-            if 'ref_to_fetch' not in source.values()[0] and source_name == 'git':
-                source[source_name]['ref_to_fetch'] = None
+            if source_name == 'git':
+                if 'ref_to_fetch' not in source.values()[0]:
+                    source[source_name]['ref_to_fetch'] = None
+                if 'archive_src_dir' not in source.values()[0]:
+                    source[source_name]['archive_src_dir'] = None
```
ghost: The previous comment applies here too, if you accept it. Both nested ifs could be removed.

lib/package.py
```diff
@@ -144,6 +144,8 @@ def _load(self):
                     source[source_name]['ref_to_fetch'] = None
                 if 'archive_src_dir' not in source.values()[0]:
                     source[source_name]['archive_src_dir'] = None
+                if 'shallow' not in source.values()[0]:
+                    source[source_name]['shallow'] = False
```
ghost: I think "shallow" should be an option to the script, not to a specific package, just like "http_proxy". It is not required by any specific package, so I see no reason to include it in yaml files.

mopsfelder: Perhaps we can have `--git-shallow` (or whatever option name makes sense here) implemented in `host_os.py` script to use git shallow clones.

Or we can implement a `git_args` parameter in `config.yaml` file and add git shallow options there; similar to what we did for `mock_args`.

alanoe: OH> I think "shallow" should be an option to the script, not to a specific package, just like "http_proxy". It is not required by any specific package, so I see no reason to include it in yaml files.

Using shallow indeed is not a requirement, but it is a *package-specific* optimization. Cloning shallowly only makes a difference in big git repositories. Unless we are *sure* we'll make this optimization for all packages, it should be package specific IMHO. And if we are sure we'll make it for all packages, I don't see the point of having an option for it.

lib/package.py

alanoe:
I've discussed this with Olav on IRC and we agreed that all git repos should be shallowly cloned by default. We see no advantage in allowing to override this per package, so I'll test if it works to clone all repos shallowly; if it works, I'll remove the per-package option.

mopsfelder:
I agree with using shallow clones by default.  However, I think code should provide users with a mechanism not to use this behaviour, if they don't want to.

lib/repository.py
```diff
@@ -51,33 +51,29 @@ def get_git_repository(remote_repo_url, parent_dir_path):
...
-                                 "http.proxy='{}'".format(proxy),
-                                 'clone',
-                                 remote_repo_url,
-                                 repo_path])
-                return GitRepository(repo_path)
-            else:
-                return super(GitRepository, cls).clone_from(
-                    remote_repo_url, repo_path, *args, **kwargs)
+            return super(GitRepository, cls).clone_from(
+                remote_repo_url, repo_path, *args, **kwargs)
```
mopsfelder, line 76: @alanoe Is proxy support still working with this change?

I recall @snowcrshd had some issues with making GitPython to work with proxy so he had to implement some hacks.

ghost: I'm not sure how passing the `config` in the options dictionary work.

If it does the same thing that my previous implementation, that's great!

ghost:
We've just realized using shallow clone is a bad idea. Git currently only supports shallow cloning by specifying a branch. This would clone the repository with only the last n commits of the specified branch. Since the packages are checked out by commit ID, there's no way to guarantee the commit we're interested in will be among those in the top of the branch that was cloned.

ghost:
@olavphibm maybe we could add an attribute to the YAML syntax indicating that we should use a shallow clone. In this case, we could default to the branch name instead of the commit id.

What do you guys think?

ghost:
@snowcrshd 
When we select a commit from the versions repo to build, the resulting packages should be the same every time the script is executed. Since we have no control over the package repositories we build, we cannot make sure the branches specified will have the same tip for executions at different times. So I don't think we could default to the branch name.

alanoe:
gcc mirror git repository is not identical to the svn repository, so we'll keep using the svn repository. As these new options are not needed, I've closed this pull request.

PR #149, opened 2017-01-16
Add package name to build directory path

ghost:


ghost:
This will require an update to the infrastructure repository: changing the logs archive path.

lib/mockbuilder.py
```diff
@@ -142,7 +142,7 @@ def _install_external_dependencies(self, package):
     def _create_build_directory(self, package):
         self.build_dir = os.path.join(
             os.getcwd(), 'build',
-            datetime.datetime.now().strftime('%Y_%m_%d_%H_%M_%S'))
+            datetime.datetime.now().isoformat())
```
farosas: The long commit message does not match the change.

farosas (changes requested):

lib/mockbuilder.py
ghost: Fixed.

farosas (approved):

mopsfelder (approved):

PR #150, opened 2017-01-17
Fix backwards compatibility with CentOS 7.2

ghost:
The scripts were broken when trying to build earlier versions because
if the config was changed to use CentOS 7.2, the Mock configuration file
would not be found and the version was not in the supported list.


farosas (approved):

olavphibm:
start build

PR #152, opened 2017-01-18
Add support to retry on timeout and/or failure when downloading package sources

ghost:


ghost:
I'll probably split the `if proxy` change into another commit.

lib/package_source.py
```diff
@@ -19,20 +21,32 @@ def _hg_download(source, directory):
...
+    if proxy:
+        command += '--config http_proxy.host="{}" '.format(proxy)
+    command += "--ssh '/usr/bin/env ssh -o ConnectTimeout={}'"
+
+    def _clone_repository(timeout):
+        cmd = command.format(timeout)
+        utils.run_command(cmd)
+
+    def _is_timeout_error(e):
+        if not isinstance(e, exception.SubprocessError): return False
```
ghost: Compound statements are discouraged by PEP8.

lib/exception.py
```diff
@@ -78,3 +78,9 @@ class SubprocessError(BaseException):
         "stdout: %(stdout)s, stderr: %(stderr)s")
     # Subclass errors are in the form 0b0100xxx
     error_code = 32
+
+
+class TimeoutError(BaseException):
+    DEFAULT_MESSAGE = ("Timeout failure on %(func_name)s after %(num_attempts)s attempts. Initial timeout: "
```
ghost: This line should be broken.

lib/package_source.py
```diff
@@ -19,20 +21,32 @@ def _hg_download(source, directory):
...
+        return ('timed out' in e.stdout or 'timed out' in e.stdout)
```
ghost: Both sides of the or are the same.

lib/package_source.py
```diff
@@ -19,20 +21,32 @@ def _hg_download(source, directory):
...
-    utils.run_command(command)
+    if proxy:
+        command += '--config http_proxy.host="{}" '.format(proxy)
+    command += "--ssh '/usr/bin/env ssh -o ConnectTimeout={}'"
+
+    def _clone_repository(timeout):
+        cmd = command.format(timeout)
+        utils.run_command(cmd)
+
+    def _is_timeout_error(e):
```
ghost: @fabianorosas, @mopsfelder and me agreed that using the same var name for exceptions everywhere would be better, and the consensus was to use `exc`. Do you agree?

lib/utils.py
```diff
@@ -16,12 +16,93 @@
...
 import subprocess
+import time
+
 
 from lib import exception
 
 LOG = logging.getLogger(__name__)
 
 
+def retry_on_error(f , error=Exception, failure_handler=None,
```
ghost: Remove the space after `f`

ghost (changes requested):

lib/package_source.py
mopsfelder: Compound statements are generally discouraged by PEP 8.

lib/package_source.py
mopsfelder: Does it make sense to have a `utils.is_timeout_error()` function?

ghost: Not sure. Timeout errors could vary in "flavour". I don't see much use in having a default checker in `utils.py`

lib/utils.py
ghost: Fixed

lib/package_source.py
ghost: Fixed

lib/package_source.py
ghost: I like`e` better (because that's what I most frequently see in python code), but if `exc` is what you agreed upon I'll just go along with that.

lib/package_source.py
ghost: Fixed

lib/exception.py
ghost: Fixed

lib/package_source.py
ghost: Fixed

ghost:
I've applied the following changes to this PR:

- Renamed exceptions from `e` to `exc`.
- Broke `TimeoutError` default message template into multiple lines.
- Fixed whitespace after `f` parameter.
- Dropped compound `if` statement.

ghost (approved):

mopsfelder (approved):
That's a nice improvement to the scripts.  Thanks!

PR #154, opened 2017-01-19
Add Subversion as a possible input of source files for a package

alanoe:


lib/repository.py
```diff
@@ -178,3 +191,55 @@ def archive(self, archive_name, commit_id, build_dir):
...
+        return os.path.basename(self.working_copy_dir)
+
+    def checkout(self, revision):
+        """
+        Check out a revision.
+        """
+        LOG.info("%(name)s: Updating svn repository"
+                 % dict(name=self.name))
+        try:
+            run_command("svn update")
```
ghost: This is missing the dir path. You could use the run_command's `cwd` attribute.

lib/repository.py
```diff
@@ -56,6 +56,19 @@ def get_git_repository(remote_repo_url, parent_dir_path):
...
 
+def get_svn_repository(remote_repo_url, repo_path):
+    """
+    Get a local subversion repository located in a directory.
+    If it does not exist, check it out from the specified URL.
+    """
+    if os.path.exists(repo_path):
+        return SvnRepository(remote_repo_url, repo_path)
+    else:
+        # TODO: setup HTTP proxy by editing ~/.subversion/servers
```
ghost, line 66: I believe the guys behind a proxy work with the kernel, which has gcc as a dependency. If that's the case, we can't leave this unfinished.

lib/repository.py
```diff
@@ -178,3 +191,55 @@ def archive(self, archive_name, commit_id, build_dir):
...
+            LOG.debug("%(name)s: Failed to update svn repository"
+                      % dict(name=self.name))
+            pass
+        else:
+            LOG.info("%(name)s: Updated svn repository" % dict(name=self.name))
+
+        LOG.info("%(name)s: Checking out revision %(revision)s"
+                 % dict(name=self.name, revision=revision))
+        try:
+            run_command("svn checkout %(repo_url)s@%(revision)s" % dict(repo_url=self.url, revision=revision))
```
ghost: This is missing the dir path.

ghost (changes requested):

lib/repository.py
```diff
@@ -23,7 +23,7 @@
 
 from lib import config
 from lib import exception
-
+from lib.utils import run_command
```
ghost: This is also being imported above in `import utils` and is being used in two different formats. Could you make it consistent and leave only one of the two imports?

lib/repository.py
alanoe: As SVN has a global config for this, IMHO this should be done before building anything instead of doing it specifically for the gcc build. We may (1) automate it and do it before building anything in host_os.py or (2) require user to do it manually and document it. 1 is better, obviously, but if the kernel team already uses our build scripts to build gcc, they already have setup this manually. The proxy manual setup in svn config files was a requirement before and remains to be it, so I do not see this as a blocker and imho it could be implemented in a future patch.

lib/repository.py
alanoe: Done

lib/repository.py
alanoe: Done

lib/repository.py
alanoe: OK, I'll do it in a separate patch.

lib/repository.py
ghost: There's no need to edit the servers file, we can set a command line option: http://stackoverflow.com/a/28503928

alanoe: Thanks for the info, I'll add it then.

My point is still valid. If a team requires using a proxy, they have never built gcc or they have already setup a proxy in their subversion configuration, as the current command used to download gcc source code (svn export) would not work without it.

ghost: It makes sense. If you prefer to leave the TODO, ok then, but let's not take too long to implement this.

olavphibm:
start build

alanoe:
start build

ghost (approved):

PR #155, opened 2017-01-20
Allow the host_os.py script to execute in parallel

ghost:
The only restriction is that two processes cannot build the same package at the same time.


lib/mockbuilder.py
```diff
@@ -41,25 +41,28 @@ def __init__(self, config):
         self.build_dir = None
         self.archive = None
         self.timestamp = datetime.datetime.now().isoformat()
+        self.common_mock_args = (
+            "%(binary)s -r %(config_file)s %(extra_args)s "
+            "--uniqueext %(suffix)s" % dict(
+                binary=MOCK_BIN, config_file=self.mock_config,
+                extra_args=self.mock_args, suffix=self.timestamp))
```
mopsfelder: What if we added these `common_mock_args` in `config.yaml` file under `mock_args` parameter?

lib/package.py
```diff
@@ -180,3 +184,31 @@ def _download_build_files(self):
...
+        resources.
+        Currently, the only resource shared among scripts executed from
+        different directories is the repository.
+        """
+        LOG.debug("Checking for lock on file {}.".format(self.lock_file_path))
+        self.lock_file = open(self.lock_file_path, "w")
+        try:
+            fcntl.lockf(self.lock_file, fcntl.LOCK_EX | fcntl.LOCK_NB)
+        except IOError as exc:
+            if exc.errno == 11:
```
mopsfelder: Where this number `11` comes from?  What if we declared a constant with a meaningful name to compare against `exc.errno`?

lib/package.py
```diff
@@ -180,3 +184,31 @@ def _download_build_files(self):
...
+        """
+        Locks the package to avoid concurrent operations on its shared
+        resources.
+        Currently, the only resource shared among scripts executed from
+        different directories is the repository.
+        """
+        LOG.debug("Checking for lock on file {}.".format(self.lock_file_path))
+        self.lock_file = open(self.lock_file_path, "w")
+        try:
+            fcntl.lockf(self.lock_file, fcntl.LOCK_EX | fcntl.LOCK_NB)
```
mopsfelder, line 214: What if `lock_file` is locked, i.e. opened by another process?  Shall we use any busy-waiting technique to wait until it can get the lock?

ghost: This check is only useful to print the log message, so the user knows why the process is hanging.
The next lock, outside the try block, will block the process until the resource is available.

lib/package.py
ghost: This message is returned when the resource is currently locked: `IOError: [Errno 11] Resource temporarily unavailable`. I've created a constant with a meaningful name.

lib/mockbuilder.py
ghost: I don't think we should merge all the arguments together. The binary path and config file are required and so I think each one of them deserves a separate parameter. I'm adding those to the config file. The mock_args are optional, so I'm leaving as is. The timestamp is generated dynamically, so a parameter for it does not make much sense.

olavphibm:
start build

lib/mockbuilder.py
```diff
@@ -33,33 +33,34 @@
 
 
 class Mock(build_system.PackageBuilder):
-    def __init__(self, config):
+    def __init__(self, mock_config):
```
alanoe: This has nothing to do with the patch scope. Mention in longer commit msg that you are renaming variables for consistency. Also, minor point: mock_config_file or mock_config_file_path would be more intuitive IMHO

lib/mockbuilder.py
```diff
@@ -33,33 +33,34 @@
...
         super(Mock, self).__init__()
-        self.mock_config = config
-        self.mock_args = CONF.get('default').get('mock_args')
+        mock_args = CONF.get('default').get('mock_args')
         self.result_dir = CONF.get('default').get('result_dir')
         self.build_dir = None
         self.archive = None
         self.timestamp = datetime.datetime.now().isoformat()
+        self.common_mock_args = (
+            "%(binary)s -r %(config_file)s %(extra_args)s" % dict(
```
alanoe: Can't the extra_args be defined in the config file so that just need to call mock binary with the config file as parameter?

alanoe: any reason to have different name for parameter name and parameter value (eg config_file and mock_config, extra_args and mock_args)? This makes code a little harder to understand

config.yaml
```diff
@@ -9,6 +9,7 @@ default:
  arch_and_endianness: 'ppc64le'
  release_notes_repo_url: "https://github.com/open-power-host-os/open-power-host-os.github.io.git"
  release_notes_repo_branch: "master"
+ mock_binary: "/usr/bin/mock"
```
alanoe, line 12: I'm not sure what is the benefit of this patch. Pls document it in commit longer msg so it is clearer.

config.yaml
```diff
@@ -10,6 +10,7 @@ default:
  release_notes_repo_url: "https://github.com/open-power-host-os/open-power-host-os.github.io.git"
  release_notes_repo_branch: "master"
  mock_binary: "/usr/bin/mock"
+ mock_config: "./extras/centOS/7/mock/epel-7-ppc64le.cfg"
```
alanoe: I do not understand this. Before, we had the ability to define a different Mock config file for each distro version. It seems to me we are losing this feature with this patch. -1 until motivation is clearer.

lib/build_manager.py
```diff
@@ -33,7 +33,8 @@ def __init__(self, packages_names, distro):
     def __call__(self):
```
alanoe, line 33: The explanation in longer commit msg is confusing. You are assuming we'll use the same directory for cloning a repository and checking out a commit for different builds? This does not need to be necessary. We can keep a table of repo_name+commit_id -> directory_path and clone to a different directory whenever commit ID changes.

lib/distro.py
```diff
@@ -57,8 +57,11 @@ def build_packages(self, packages):
         """
```
alanoe, line 57: I have not found out how to comment the commit msg, so I'm doing it here. Use "lock files" instead of "lockfiles" pls, better English

lib/package.py
alanoe: IMHO locking the source code repository when someone is using it is not an optimal solution. This only allows the simultaneous build of one package at once and only parallelizes builds of different packages. I'd prefer to use a  solution which parallelizes builds of *any* package , like using a dictionary (repo_name, commit_id) -> repo_directory_path

alanoe (changes requested):

lib/mockbuilder.py
ghost: Done.

lib/mockbuilder.py
ghost: I've removed the `mock_` prefix, since the class is already named `Mock`, and left the names and values with the same name.

ghost: It is defined in the config file, but can be overwritten by a command line parameter.

config.yaml
ghost: Done. The user may want to use an alternate Mock binary or have it installed in a different path.

config.yaml
ghost: If you look down at centos.py, you'll see that the path is actually hardcoded, regardless of the CentOS version, so we're not losing any flexibility with this. If the user wants to use a different Mock config, he can now modify the config.yaml file.

lib/distro.py
ghost: Done.

lib/build_manager.py
ghost: Yes, I'm assuming there will be only one shared repository for each package for all builds. This avoids having to clone potentially big repositories multiple times. The actual build can be executed in parallel, since after the archive is ready the repository is not used anymore. I'm adding this improvement in a separate commit.
I'm improving the commit message.

olavphibm:
start build

lib/mockbuilder.py
```diff
@@ -42,9 +42,10 @@ def __init__(self):
         self.archive = None
```
alanoe, line 41: In the longer commit msg, I'd prefer "This allows multiples processes to use Mock concurrently, creating a separate chroot for each". This makes clearer that different processes do *not* share a chroot.

lib/build_manager.py
alanoe: New commit msg is much better!

lib/distro.py
```diff
@@ -57,8 +57,11 @@ def build_packages(self, packages):
...
         self.package_builder.initialize()
         for package in packages:
+            package.lock()
```
alanoe, line 60: Do we need to lock if the package uses only the url source?

config.yaml
alanoe: Is this a requirement? Is there a concrete use case (have you ever had the need of use an alternate Mock binary or found a mock rpm package which installs it in a different path) ?

config.yaml
alanoe: As we talked personally, there should be one Mock config file should be defined per distro and distro version. I noticed that the previous code (incorrectly) always used the same Mock config file for any distro version, but at least it used one different Mock config file for each distro. This patch allows one Mock config for all builds, reducing build script flexibility. Pls use subattribs in the yaml for 'centOS' and '7' as we have in the packages metadata (yaml) file.

alanoe (changes requested):

lib/distro.py
```diff
@@ -57,10 +57,10 @@ def build_packages(self, packages):
         """
         self.package_builder.initialize()
         for package in packages:
+            # The package unlock is done inside package_builder.build()
             package.lock()
             package.download_files(recurse=False)
             self.package_builder.build(package)
-            package.unlock()
```
alanoe: It is simpler to understand the locking mechanism if lock() and unlock() are in the same level. I suggest moving the source code packaging to outside package_builder.build().

lib/mockbuilder.py
ghost: Done.

config.yaml
ghost: Some time ago, we had issues with the mock binary requiring root because we called the command `mock` instead of `/usr/bin/mock`, which lead to a different binary in some machines.
I remember someone commenting that it would be nice to have this configurable, so I think its harmless and maybe useful for someone to have this flexibility.

config.yaml
ghost: Done.

lib/distro.py
ghost: I agree with that, but I don't want to break the build function in too many steps and make the Mock class' usability harder. I've decided to externalize the `Mock._prepare()` function, with the small drawback that the build_files copying will be included in the lock.

lib/distro.py
ghost: No. I've added a commit to disable locking for those packages.

mopsfelder (approved):

lib/mockbuilder.py
```diff
@@ -57,7 +57,6 @@ def initialize(self):
 
     def build(self, package):
         LOG.info("%s: Starting build process" % package.name)
-        self._prepare(package)
```
alanoe, line 57: why this line was removed in this patch? It seems to me it should have been removed in the previous patch

ghost: That was a mistake indeed. The PR is already merged though, so there's nothing I can do.

PR #158, opened 2017-01-25
Add scripts to validate packages metadata (yaml) files and rpm spec f…

alanoe:
…iles

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -13,6 +13,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import fnmatch
 import logging
 import os
 import subprocess
@@ -137,3 +138,45 @@ def create_directory(directory):
     """
     if not os.path.isdir(directory):
         os.makedirs(directory)
+
+
+def is_package_installed(package_name):
+    """
+    Checks if a RPM package is installed
+
+    Args:
+        package_name (str): package name
+
+    Returns:
+        bool: if RPM package is installed
+    """
+
+    cmd = "rpm -q %s" % package_name
+    try:
+        run_command(cmd, shell=True)
+    except exception.SubprocessError as e:
+        # rpm returns 1 when search string is not found and other non-zero values
+        # if an error occurred
+        #pylint: disable=no-member
+        if e.returncode == 1:
+            return False
+        else:
+            raise
+
+    return True
+
+
+def recursive_glob(directory, pattern):
+    """
+    Find all files matching a pattern according to the rules used by Linux shell
+
+    Args:
+        directory (str): searched directory
+        pattern (str): glob pattern
+    """
+
+    matches = []
+    for root, _, filenames in os.walk(directory):
+        for filename in fnmatch.filter(filenames, pattern):
+            matches.append(os.path.join(root, filename))
+    return matches
--- a/validate_rpm_specs.py
+++ b/validate_rpm_specs.py
@@ -1,0 +1,87 @@
+#!/usr/bin/env python
+# Copyright (C) IBM Corp. 2016.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import argparse
+import sys
+
+from lib import exception
+from lib.utils import is_package_installed
+from lib.utils import recursive_glob
+from lib.utils import run_command
+
+def validate_rpm_spec(spec_file_path):
+    """
+    Validate RPM specification file
+
+    Args:
+        spec_file_path (str): RPM specification file path
+
+    Returns:
+        bool: if RPM specification file is valid
+    """
+
+    try:
+        run_command("rpmlint -f .rpmlint -v %s" % spec_file_path)
+    except exception.SubprocessError as e:
+        #pylint: disable=no-member
+        print("validation of RPM specification file %s failed, output: %s" % (spec_file_path, e.stdout))
+        return False
+
+    return True
+
+
+def validate_rpm_specs(base_dir):
+    """
+    Validate specification files of rpm packages in a base directory
+
+    Args:
+        base_dir (str): base directory path
+
+    Returns:
+        bool: if RPM specification files are valid
+    """
+
+    files = recursive_glob(base_dir, "*.spec")
+    valid = True
+    for _file in files:
+        if not validate_rpm_spec(_file):
+            valid = False
+
+    return valid
+
+
+def parse_cli_options():
+    """
+    Parse CLI options
+
+    Returns:
+        Namespace: CLI options. Valid attributes: rpm_specs_base_dir
+    """
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-d', '--rpm-specs-base-dir', dest='rpm_specs_base_dir',
+                        required=True, help='RPM specification files base directory path')
+    args = parser.parse_args()
+    return args
+
+
+if __name__ == '__main__':
+    if not is_package_installed('rpmlint'):
+        print("rpmlint package should be installed before running this script")
+        sys.exit(1)
+    args = parse_cli_options()
+    if not validate_rpm_specs(args.rpm_specs_base_dir):
+        sys.exit(2)
--- a/validate_yamls.py
+++ b/validate_yamls.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python
+# Copyright (C) IBM Corp. 2016.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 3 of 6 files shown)


validate_rpm_specs.py
```diff
@@ -0,0 +1,53 @@
...
+import sys
+
+from lib import exception
+from lib.utils import is_package_installed
+from lib.utils import run_command
+
+def validate_rpm_spec(spec_file_path):
+
+    try:
+        run_command("rpmlint -v %s" % spec_file_path)
```
ghost: Unrelated to this patch: we should probably have a file to keep those dev package requirements, like rpm_requirements.txt

validate_rpm_specs.py
```diff
@@ -0,0 +1,53 @@
...
+    except exception.SubprocessError as e:
+        #pylint: disable=no-member
+        print("validation of rpm spec file %s failed, output: %s" % (spec_file_path, e.stdout))
+        return
+
+def validate_rpm_specs(base_dir):
+    """
+    Validate specification files of rpm packages built by Open Power Host OS
```
ghost: OpenPOWER is not separated.

ghost (commented):
Those files look very similar, why not have a single `validate_files.py` script with options or subcommands for spec files, yaml files and maybe even python files?

lib/utils.py
```diff
@@ -45,3 +45,23 @@ def run_command(cmd, **kwargs):
                                         stdout=output, stderr=error_output)
 
     return output
+
+
+def is_package_installed(package_name):
+    """
+    Checks if a RPM package is installed
```
mopsfelder, line 145: What about `Returns True if an RPM package is installed, False otherwise.`?

validate_rpm_specs.py
mopsfelder: I think we should avoid branding the code.

validate_rpm_specs.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+    packages = os.listdir(base_dir)
+
+    for package in packages:
+        files = glob.glob("{base_dir}/{package_name}/*/*/{package_name}.spec".format(
```
mopsfelder: Since `base_dir` is passed, I think the code should recursively search by all `*.spec` files under `base_dir` directory and validate them all.  Thus, if directory structure changes in the future, that won't break this function.

validate_rpm_specs.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Parse CLI options
+
+    Returns:
+        Namespace: CLI options. Valid attributes: packages_metadata_base_dir
+    """
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-p', '--packages-metadata-base-dir', dest='packages_metadata_base_dir',
+                        required=True, help='packages metadata base directory path')
```
mopsfelder: I think `--specs-dir` would reflect more the meaning of the parameter since it's the base directory of where spec files reside.  That would also make `validate_rpm_specs.py` a more generic tool, not tailored only for Host OS packages.

validate_yamls.py
```diff
@@ -0,0 +1,53 @@
...
+    try:
+        run_command("yamllint %s" % package_metadata_file_path)
+    except exception.SubprocessError as e:
+        #pylint: disable=no-member
+        print("validation of package metadata file %s failed, output: %s" % (package_metadata_file_path, e.stdout))
+        return
+
+def validate_packages_metadata(base_dir):
+    """
+    Validate specification files of rpm packages built by Open Power Host OS
```
mopsfelder: Same here.  Let's try to avoid branded code or comments.

validate_yamls.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+    packages = os.listdir(base_dir)
+
+    for package in packages:
+        files = glob.glob("{base_dir}/{package_name}/{package_name}.yaml".format(
```
mopsfelder: Same here.  Since `base_dir` is given, code can walk through this directory recursively and search for all `*.yaml` files to validate.

validate_yamls.py
```diff
@@ -0,0 +1,53 @@
...
+def validate_package_metadata(package_metadata_file_path):
+
+    try:
+        run_command("yamllint %s" % package_metadata_file_path)
+    except exception.SubprocessError as e:
+        #pylint: disable=no-member
+        print("validation of package metadata file %s failed, output: %s" % (package_metadata_file_path, e.stdout))
+        return
+
+def validate_packages_metadata(base_dir):
```
mopsfelder: I think a more generic name for the function may be welcome, e.g. `validate_yaml_file()`.  Similarly to the above `validate_rpm_spec()`.

validate_yamls.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Parse CLI options
+
+    Returns:
+        Namespace: CLI options. Valid attributes: packages_metadata_base_dir
+    """
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-p', '--packages-metadata-base-dir', dest='packages_metadata_base_dir',
+                        required=True, help='packages metadata base directory path')
```
mopsfelder: Same here.  Perhaps a `--yamls-dir` or just `--base-dir` allows the tool to be more generic.

validate_yamls.py
```diff
@@ -0,0 +1,53 @@
...
+import argparse
+import glob
+import os
+import sys
+
+from lib import exception
+from lib.utils import is_package_installed
+from lib.utils import run_command
+
+def validate_package_metadata(package_metadata_file_path):
```
mopsfelder: Perhaps a more generic name for this function may be welcome, e.g. `validate_yaml_files()`.

validate_rpm_specs.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python
+
```
farosas: This file is missing the Copyright statement

farosas (changes requested):

ghost:
The script exits with code 0 even when there are erros in the yaml files.
The functions `validate_package_metadata` and `validate_packages_metadata` should either return a boolean or raise an exception, so the script can detect it and exit with an error code.

validate_rpm_specs.py
alanoe: Done

lib/utils.py
alanoe: I've added Google-style param and return value doc to this function. I suggest we do the same for the rest of the code incrementally.

validate_rpm_specs.py
alanoe: glob() in Python 2.x does not support recursion, but I will implement a function which does it.

validate_rpm_specs.py
alanoe: Done

validate_yamls.py
alanoe: Done

validate_yamls.py
alanoe: Done

validate_yamls.py
alanoe: As the package metadata yaml file format is ours, we could use function this to validate our syntax too. Anyway, as were are not doing this for now, I agree with your suggestion.

validate_yamls.py
alanoe: Done

validate_rpm_specs.py
alanoe: Question: do we need to add the Copyright statement to *every* file ? Creating a text file with it in the root directory is not enough?

alanoe:
OH> Those files look very similar, why not have a single validate_files.py script with options or subcommands for spec files, yaml files and maybe even python files?

It can be done, just not sure if it's worth it the complexity. It may be worth it if we'll often desire to run all those validations at once manually. 

OH> The script exits with code 0 even when there are erros in the yaml files.
The functions validate_package_metadata and validate_packages_metadata should either return a boolean or raise an exception, so the script can detect it and exit with an error code.

Done

validate_rpm_specs.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    Parse CLI options
+
+    Returns:
+        Namespace: CLI options. Valid attributes: rpm_specs_base_dir
+    """
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-d', '--rpm-specs-base-dir', dest='rpm_specs_base_dir',
+                        required=True, help='RPM specification files base directory path')
```
ghost, line 76: Would be easier to have `default='.'`

validate_yamls.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    Parse CLI options
+
+    Returns:
+        Namespace: CLI options. Valid attributes: yamls_base_dir
+    """
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-d', '--yamls-base-dir', dest='yamls_base_dir',
+                        required=True, help='YAML files base directory path')
```
ghost, line 76: Would be easier to have `default='.'`

validate_rpm_specs.py
alanoe: I usually run this from builds root directory, so I have to use the parameter anyway. I can set the default to the current directory if you think it will be useful.

validate_yamls.py
alanoe: Same as above

validate_rpm_specs.py
ghost: I think a default would be nice. Maybe it's better to use 'versions' instead of '.'

alanoe: It would be, but I'm not sure which one to use. We devs would like to use ../versions in our workstations, but the correct path would be ./versions in the build system. And the latter is prone to change to /var/lib/host-os, IIRC.

ghost: You're right. Just leave it without a default then. Maybe we could read it from the config in the future, if we ever decide to use the config.py from this file.

ghost (approved):

PR #159, opened 2017-01-25
Fix upgrade-versions command

ghost:


tools/upgrade_versions.py
```diff
@@ -114,13 +114,17 @@ def update(self, user_name, user_email):
 
     def _bump_release(self, pkg, log=None, user_name=None, user_email=None):
         LOG.info("%s: Bumping release" % self.pkg)
+        if log is None:
+            log = []
 
-        if self.pkg.commit_id:
+        old_commit_id = self.pkg.sources[0]["git"]["commit_id"]
+        if old_commit_id:
+            new_commit_id = pkg.sources[0]["git"]["commit_id"]
```
alanoe: I'd prefer to put pkg.sources[0]["git"] in a new_source variable instead of duplicating it here and below, but up to you

tools/upgrade_versions.py
```diff
@@ -114,13 +114,17 @@ def update(self, user_name, user_email):
 
     def _bump_release(self, pkg, log=None, user_name=None, user_email=None):
```
alanoe: Does it make sense to accept a list here? I'd accept a str instead and create the log list with the str as the list first element. This would remove the need of making changelog variable above a list, which does not make sense IMHO

alanoe (changes requested):

tools/upgrade_versions.py
ghost: Using a list isn't useful for our current usage, I'm applying your suggestion and renaming the variable to something better.

tools/upgrade_versions.py
ghost: Done.

PR #161, opened 2017-01-26
Reorganize mock config directories and files

alanoe:


ghost (commented):
This requires a modification in the infrastructure repository too, because we replace the main centos mirror configured in the cfg file with our internal one: https://github.com/open-power-host-os/infrastructure/blob/master/jenkins_jobs/build_host_os/script.sh

It will also conflict with https://github.com/open-power-host-os/builds/pull/155, which adds a mock_config in the config.yaml file.

ghost (commented):
What about the consistency with the versions repository?

alanoe:
Consistency with versions git repository is done in https://github.com/open-power-host-os/versions/pull/103

ghost (approved):
The test failed because it depends on https://github.com/open-power-host-os/infrastructure/pull/18. I've tested it in my environment and it's ok to merge.

PR #162, opened 2017-01-27
Retry SVN repository checkout on failure

ghost:


lib/package_source.py
```diff
@@ -88,8 +88,12 @@ def _svn_download(source, directory):
         raise ValueError('invalid subversion source dict: missing both `commit_id` '
                          'and `branch`')
 
-    repo = repository.get_svn_repository(svn_source['src'],
-                                         directory)
+    def _download_repository():
+        return repository.get_svn_repository(svn_source['src'], directory)
```
alanoe, line 92: I personally dislike nested methods, but ok...

alanoe (approved):

lib/package_source.py
ghost: This is a single use method. Defining it outside would require adding parameters, which are not currently supported by the rety function and would overcomplicate it, in my opinion. It is also consistent with the retry calls for the other source types.

PR #163, opened 2017-01-27
Update commit ID in rpm specification files when updating commit ID i…

alanoe:
…n yaml files in upgrade_versions.py

### DIFF
--- a/lib/rpm_package.py
+++ b/lib/rpm_package.py
@@ -103,7 +103,7 @@ class SpecFile(object):
 
         self.write_content()
 
-    def bump_release(self, change_log_lines, user_name, user_email):
+    def update_release_and_change_log(self, change_log_lines, user_name, user_email):
         comment = "\n".join(['- ' + l for l in change_log_lines])
         user_string = "%(user_name)s <%(user_email)s>" % locals()
         cmd = "rpmdev-bumpspec -c '%s' -u '%s' %s" % (
--- a/lib/rpm_package.py
+++ b/lib/rpm_package.py
@@ -119,6 +119,9 @@ class SpecFile(object):
         LOG.info("Updated '%s' prerelease tag to: %s"
                  % (self.path, new_prerelease))
 
+    def update_commit_id(self, old_commit_id, new_commit_id):
+        utils.replace_str_in_file(self.path, old_commit_id, new_commit_id)
+
     def _replace_macro_definition(self, macro_name, replacement):
         """
         Updates the file content cache, replacing the macro value.
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -137,3 +137,14 @@ def create_directory(directory):
     """
     if not os.path.isdir(directory):
         os.makedirs(directory)
+
+
+def replace_str_in_file(file_path, search, replacement):
+
+    lines = []
+    with file(file_path, "r") as f:
+        lines = f.readlines()
+    with file(file_path, "w") as f:
+        for line in lines:
+            line = line.replace(search, replacement)
+            f.write(line)
--- a/tools/upgrade_versions.py
+++ b/tools/upgrade_versions.py
@@ -26,6 +26,7 @@ from lib import exception
 from lib import packages_manager
 from lib import repository
 from lib import rpm_package
+from lib.utils import replace_str_in_file
 from lib.versions_repository import setup_versions_repository
 
 
@@ -49,16 +50,6 @@ PACKAGES = [
 
 # prerelease strings supported as last element in the version regex
 PRERELEASE_TERMS = ['rc']
-
-
-def _sed_yaml_descriptor(yamlfile, old_commit, new_commit):
-    lines = []
-    with file(yamlfile, "r") as f:
-        lines = f.readlines()
-    with file(yamlfile, "w") as f:
-        for line in lines:
-            line = line.replace(old_commit, new_commit)
-            f.write(line)
 
 
 def _get_git_log(repo, since_id):
@@ -130,8 +121,9 @@ class Version(object):
                 self.pkg.name, old_commit_id, new_commit_id))
             change_log_lines += _get_git_log(
                 new_source["repo"], old_commit_id)
-            _sed_yaml_descriptor(
+            replace_str_in_file(
                 self.pkg.package_file, old_commit_id, new_commit_id)
+            pkg.spec_file.replace_commit_id(old_commit_id, new_commit_id)
 
         if change_log_lines:
             assert user_name is not None
--- a/tools/upgrade_versions.py
+++ b/tools/upgrade_versions.py
@@ -104,9 +104,9 @@ class Version(object):
                 (self.pkg.version, self._repo_version))
 
         pkg.spec_file.update_prerelease_tag(self._repo_prerelease)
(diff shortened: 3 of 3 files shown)


tools/upgrade_versions.py
```diff
@@ -61,6 +61,16 @@ def _sed_yaml_descriptor(yamlfile, old_commit, new_commit):
...
 
 
+def _sed_rpm_spec(specfile, old_commit, new_commit):
+    lines = []
+    with file(specfile, "r") as f:
+        lines = f.readlines()
+    with file(specfile, "w") as f:
+        for line in lines:
+            line = line.replace(old_commit, new_commit)
+            f.write(line)
```
ghost: This function is identical to the one above, you could just change its variables to have more generic names.

tools/upgrade_versions.py
```diff
@@ -132,6 +142,8 @@ def _bump_release(self, pkg, change_log_header=None, user_name=None,
                 new_source["repo"], old_commit_id)
             _sed_yaml_descriptor(
                 self.pkg.package_file, old_commit_id, new_commit_id)
+            _sed_rpm_spec(
+                self.pkg.package_file, old_commit_id, new_commit_id)
```
ghost: You're passing the yaml file to both functions.

ghost (changes requested):
Please try to keep short commit messages shorter than 72 characters, so they fit in the PR and commit titles.

mopsfelder:
Perhaps we can have a more generic function, say `lib.utils.inplace_replace()`, that can be used in `_sed_yaml_descriptor()` and `_sed_rpm_spec()`.  For example:

```python
import fileinput
def inplace_replace(files, pattern, replace):
  for line in fileinput.input(files=files, inplace=True):
    print(line.replace(pattern, replace))

def _sed_yaml_descriptor(yamlfile, old_commit, new_commit):
  inplace_replace(file=yamlfile, old_commit, new_commit)

def _sed_rpm_spec(specfile, old_commit, new_commit):
  inplace_replace(file=specfile, old_commit, new_commit)
```

ghost:
I like @mopsfelder's `inplace_replace` function, I just don't see a reason to keep the identical wrapper functions.

tools/upgrade_versions.py

tools/upgrade_versions.py

alanoe:
OH> Please try to keep short commit messages shorter than 72 characters, so they fit in the PR and commit titles.

We already have layout rules for Python code and commit message (assuming we have inherited the ones we used in PowerKVM), do we need to add an extra one? I personally always prefer a longer and clearer description than a non-intuitive one. It makes easier to anyone inside or outside the team to understand what the patch does without reading its content.

OH> Perhaps we can have a more generic function, say lib.utils.inplace_replace(), that can be used in _sed_yaml_descriptor() and _sed_rpm_spec()

Done.  I just used a more intuitive name. I agree with Olav that the wrapper functions are useless.

ghost:
Most sources that I've found (including git documentation: https://git-scm.com/docs/git-commit#_discussion) recommend writing the short commit message with at most 50 characters. I personally find it hard to do this with many commits, so I'm just suggesting we keep a maximum of 72 to avoid breaking readability in GitHub like this:
```
Update commit ID in rpm specification files when updating commit ID i…

…n yaml files
```
This line break also appears in the PR title which is then used as the default commit message when merging the PR.

The longer commit message can be used to explain the commit in detail.

tools/upgrade_versions.py
```diff
@@ -104,9 +104,9 @@ def update(self, user_name, user_email):
                 (self.pkg.version, self._repo_version))
 
         pkg.spec_file.update_prerelease_tag(self._repo_prerelease)
-        self._bump_release(pkg, change_log_header, user_name, user_email)
+        self._update_release_and_change_log(pkg, change_log_header, user_name, user_email)
 
-    def _bump_release(self, pkg, change_log_header=None, user_name=None,
+    def _update_release_and_change_log(self, pkg, change_log_header=None, user_name=None,
                       user_email=None):
```
ghost, line 110: Indentation became wrong.

lib/rpm_package.py
```diff
@@ -103,7 +103,7 @@ def update_version(self, new_version):
 
         self.write_content()
 
-    def bump_release(self, change_log_lines, user_name, user_email):
+    def update_release_and_change_log(self, change_log_lines, user_name, user_email):
```
farosas, line 106: Historical information: 'bump' is not the same as 'update' because bump is usually used to mean 'increment'. I used 'update' because the version might be totally wrong since we don't have control over what the package owners do.

Regardless, I don't see how bumping the release would not cause the change log to be updated or how updating the change log would not cause the release to be bumped.

tools/upgrade_versions.py
```diff
@@ -128,7 +128,7 @@ def _bump_release(self, pkg, change_log_header=None, user_name=None,
         if change_log_lines:
             assert user_name is not None
             assert user_email is not None
-            pkg.spec_file.bump_release(change_log_lines, user_name, user_email)
+            pkg.spec_file.update_release_and_change_log(change_log_lines, user_name, user_email)
```
ghost, line 131: Line is too long.

ghost (approved):

lib/utils.py
```diff
@@ -126,3 +126,14 @@ def run_command(cmd, **kwargs):
...
+
+def replace_str_in_file(file_path, search, replacement):
+
+    lines = []
+    with file(file_path, "r") as f:
+        lines = f.readlines()
+    with file(file_path, "w") as f:
+        for line in lines:
+            line = line.replace(search, replacement)
+            f.write(line)
```
ghost, line 150: What about @mopsfelder's suggestion of using fileinput here? It simplifies this function.

alanoe:
OH> I'm just suggesting we keep a maximum of 72 to avoid breaking readability in GitHub like this

I have understood the motivation, but I think this issue is *very* minor because you'll need to open the PR to look at the patch anyway and then you'll see the full commit msg. Besides that, I think describing the patch clearly is much more important than having a concise commit msg. Anyway, for most cases, using the longer commit msg to describe the patch better works. I just wished we spent more time discussing algorithms, design and performance than code/doc layout in code reviews.

lib/utils.py
alanoe: Using fileinput reduces code size, but I personally dislike the print() in it. I had to open Python std lib doc to understand that "the file is moved to a backup file and standard output is directed to the input file" in it. To me, the code above is simpler to understand. But if you guys prefer using fileinput, I can use it.

ghost: I see your point. I'm ok with leaving it like this then.

lib/rpm_package.py
alanoe: FR> Regardless, I don't see how bumping the release would not cause the change log to be updated or how updating the change log would not cause the release to be bumped.

If the rpm package does not keep a change log, you could simply increment the release number. Also, it is not obvious to a beginner that a rpm spec even has a change log. Several other file descriptors (eg XML/JSON/YAML files, such as our package metadata yaml file) do not. I always recommend not making assumptions and keeping things explicit.

PR #164, opened 2017-01-27
Fix parsing of arguments for the build-iso command

ghost:
It was getting the mock_config value from the "default" node in the
config.yaml file, because it has the same name as the attribute in the
"iso" node.

Regression introduced by commit 45c1676


mopsfelder (approved):
Thanks for catching this!

olavphibm:
start build

olavphibm:
start build

Issue #166, opened 2017-01-31
host_os.py set-env --user <user> is not set properly directory permission


gutoyr:
After execute `sudo host_os.py set-env --user <user>`, permissions under /var/log/host-os/ were not set properly. Leading to any host_os.py command fail with **IOError: [Errno 13] Permission denied: '/var/log/host-os/builds.log'**

A workaround for this issue is to run `sudo chown -R <user>:$(id -g -n <user>) /var/log/host-os` just after execute _set-env_

gutoyr:
Fixed by PR #169 
Closing issue.

PR #169, opened 2017-02-01
Fix group permission on set-env subcommand

gyribeiro:


tools/setup_environment.py
```diff
@@ -55,8 +73,8 @@ def main(user):
     log_dir = "/var/log/host-os"
     repo_dir = "/var/lib/host-os/repositories"
```
farosas: Maybe use `/var/lib/host-os` instead so that we don't need to add each directory individually should we decide to put more stuff under host-os. I remember we discussed about moving the 'versions' clone to `/var/lib/host-os` too.

gyribeiro:
start test

ghost:
start tests

ghost:
start pylint

tools/setup_environment.py
```diff
@@ -34,17 +37,35 @@ def setup_user(user):
...
+
+def setup_directory(directory, gid):
     """
     Set the directory's owners, creating it if it does not exist.
     """
     utils.create_directory(directory)
-    os.chown(directory, uid, gid)
+    set_group_permissions(gid, directory)
+    for root, dirnames, filenames in os.walk(directory):
+        for f in filenames:
```
ghost: You could write `for path in dirnames + filenames:` to avoid repeating the for loop.

ghost (approved):

tools/setup_environment.py
gyribeiro: done

farosas (approved):

gyribeiro:
This PR closes issue #166

PR #170, opened 2017-02-01
Fix 'files' attribute loading from package metadata (yaml) file

alanoe:
Regression introduced in commit a55b46d7df91d62bc0a239b28ec547b258e783b2.

### DIFF
--- a/lib/rpm_package.py
+++ b/lib/rpm_package.py
@@ -151,7 +151,7 @@ class RPM_Package(Package):
 
             # load distro files
             files = self.package_data.get('files', {}).get(
-                distro_attrib_name, {}).get(self.distro.version, {})
+                distro_attrib_name, {}).get(self.distro.version, {}) or {}
 
             default_build_files_dir_rel_path = os.path.join(
                 self.distro.lsb_name, self.distro.version, "SOURCES")


lib/rpm_package.py
```diff
@@ -151,7 +151,7 @@ def _load(self):
 
             # load distro files
             files = self.package_data.get('files', {}).get(
-                distro_attrib_name, {}).get(self.distro.version, {})
+                distro_attrib_name, {}).get(self.distro.version, {}) or {}
```
ghost, line 154: I don't get where this is needed. In cases where the `files` attribute of the yaml file is empty? Does this happen for any package?

alanoe: OH> I don't get where this is needed. 

In the case where we have files/CentOS/7 attributes defined, but nothing below them. In this case, files would be assigned None and files.get('build_files') right below this line would fail.

This happens in some yaml files (eg qemu) after PR versions/#89. We could delete the 'files' attrib in this case in that PR, but nevertheless this change is welcome IMHO.

ghost: Got it. I agree in merging this as a precaution, but I think it's better to delete the empty values still.

ghost (approved):

PR #173, opened 2017-02-01
Modify upgrade-versions to allow its usage without push-repo parameters

lisibann:


tools/upgrade_versions.py
```diff
@@ -224,13 +224,24 @@ def run(CONF):
     committer_name = CONF.get('default').get('committer_name')
     committer_email = CONF.get('default').get('committer_email')
 
-    REQUIRED_PARAMETERS = ["push_repo_url", "push_repo_branch",
-                           "committer_name", "committer_email"]
+    REQUIRED_PARAMETERS = ["committer_name", "committer_email"]
+                           
+    REQUIRED_PUSH_REPO_PARAMETERS = ["push_repo_url", "push_repo_branch"]
```
ghost, line 229: I'd remove the `REQUIRED_` prefix, since those parameters are now optional.

tools/upgrade_versions.py
```diff
@@ -224,13 +224,24 @@ def run(CONF):
...
+
     for parameter in REQUIRED_PARAMETERS:
         if CONF.get('default').get(parameter) is None:
             LOG.error("Parameter '%s' is required", parameter)
             return 1
 
+    push_repo = True
```
ghost, line 236: What if we got this boolean from the config? This way, the user would have to explicitly disable pushing (e.g. by passing --local-update) and we could just concatenate the list `PUSH_REPO_PARAMETERS` into `REQUIRED_PARAMETERS` if it's `True`, removing the second for.

tools/upgrade_versions.py
```diff
@@ -243,6 +254,7 @@ def run(CONF):
         pkg_version.update(committer_name, committer_email)
         pkg.unlock()
 
-    release_date = datetime.today().date().isoformat()
-    push_new_versions(versions_repo, release_date, push_repo_url,
-                      push_repo_branch, committer_name, committer_email)
+    if push_repo:
+        release_date = datetime.today().date().isoformat()
+        push_new_versions(versions_repo, release_date, push_repo_url,
+                          push_repo_branch, committer_name, committer_email)
```
ghost, line 260: We still want a local commit, so either we put this if inside the `push_new_versions` function or split into two, e.g. `create_commit` and `push_new_versions`. I think the latter will look better.
This local commit will then have to be passed to the `build_host_os` job in the nightly job.

lib/config.py
```diff
@@ -88,7 +88,7 @@
     ('release-notes', 'Create release notes',
         [RELEASE_NOTES_ARGS, PUSH_REPO_ARGS, BUILD_REPO_ARGS]),
     ('upgrade-versions', 'Upgrade packages versions',
-        [PUSH_REPO_ARGS, BUILD_REPO_ARGS]),
```
ghost, line 91: Please explain the motivation in the commit longer message. I believe you want to select the packages that are being upgraded?

ghost (changes requested):

alanoe:
It seems that I and Lisiane have developed the same feature by accident. My PR #174 handles the same problem and already implements most changes suggested by Olav here. I suggest we use it and close this one.

ghost:
This is already being implemented in https://github.com/open-power-host-os/builds/pull/174

PR #174, opened 2017-02-02
Update commit id in rpm specs when auto updating yamls

alanoe:


lib/rpm_package.py
```diff
@@ -119,6 +119,32 @@ def update_prerelease_tag(self, new_prerelease):
         LOG.info("Updated '%s' prerelease tag to: %s"
                  % (self.path, new_prerelease))
 
+    def update_commit_id(self, old_commit_id, new_commit_id):
+
+        # change log may contain old commit IDs and we do not want to replace them
+
+        # read up to change log
+        lines = []
+        change_log_tag = "%changelog"
```
ghost: This is a constant. Use uppercase.

tools/upgrade_versions.py
```diff
@@ -26,6 +26,7 @@
 from lib import packages_manager
 from lib import repository
 from lib import rpm_package
+from lib.utils import replace_str_in_file
```
ghost, line 29: This move is not mentioned in the commit message and could be separated to another commit.

tools/upgrade_versions.py
```diff
@@ -104,9 +104,9 @@ def update(self, user_name, user_email):
                 (self.pkg.version, self._repo_version))
 
         pkg.spec_file.update_prerelease_tag(self._repo_prerelease)
-        self._bump_release(pkg, change_log_header, user_name, user_email)
+        self._update_release_and_change_log(pkg, change_log_header, user_name, user_email)
 
-    def _bump_release(self, pkg, change_log_header=None, user_name=None,
+    def _update_release_and_change_log(self, pkg, change_log_header=None, user_name=None,
                       user_email=None):
```
ghost: Indentation is not following PEP8.

tools/upgrade_versions.py
```diff
@@ -128,7 +128,7 @@ def _bump_release(self, pkg, change_log_header=None, user_name=None,
         if change_log_lines:
             assert user_name is not None
             assert user_email is not None
-            pkg.spec_file.bump_release(change_log_lines, user_name, user_email)
+            pkg.spec_file.update_release_and_change_log(change_log_lines, user_name, user_email)
```
ghost: Line is too long.

tools/upgrade_versions.py
```diff
@@ -173,35 +173,36 @@ def _read_version_from_repo(self, repo_path):
             raise exception.PackageError(msg)
 
 
-def push_new_versions(versions_repo, release_date, versions_repo_push_url,
-                      versions_repo_push_branch, committer_name,
-                      committer_email):
+def commit_and_push_new_versions(versions_repo, release_date, versions_repo_push_url,
+                      versions_repo_push_branch, updater_name,
+                      updater_email, push):
```
ghost: Indentation is not following PEP8.

lib/config.py
```diff
@@ -62,15 +62,19 @@
...
         dict(help='URL of the repository used for pushing'),
     ('--push-repo-branch',):
         dict(help='Branch of the repository used for pushing',
              default='master'),
-    ('--committer-name',):
-        dict(help='Name used when creating a commit and bumping spec files'),
-    ('--committer-email',):
-        dict(help='Email used when creating a commit and bumping spec files'),
+    ('--updater-name',):
+        dict(help='Name used when updating files'),
```
ghost: I think it's better to specify what we are updating.
For example: `Name used when updating spec files and committing changes`
or `Name used to update spec files changelogs and create git commit`

tools/upgrade_versions.py
```diff
@@ -232,9 +238,11 @@ def run(CONF):
...
         pkg.unlock()
 
-    release_date = datetime.today().date().isoformat()
-    push_new_versions(versions_repo, release_date, push_repo_url,
-                      push_repo_branch, committer_name, committer_email)
+    if commit_updates:
+        release_date = datetime.today().date().isoformat()
+        commit_and_push_new_versions(versions_repo, release_date, push_repo_url,
+                                     push_repo_branch, updater_name, updater_email,
+                                     push_updates)
```
ghost: I think it would be better to divide this function in two to reduce the number of parameters. I suggest:
`commit_new_versions(versions_repo, release_date, updater_name, updater_email)`
`push_head_commit(versions_repo, push_repo_url, push_repo_branch)`

ghost (changes requested):

lib/config.py
```diff
@@ -62,15 +62,19 @@
         dict(help='Branch of the release notes repository to checkout'),
 }
 PUSH_REPO_ARGS = {
+    ('--commit-updates',):
+        dict(help='Commit file updates to local repository'),
+    ('--push-updates',):
+        dict(help='Push file updates to remote repository'),
```
ghost: Those parameters will require a string argument, and the resulting variable's value will always evaluate to `True`.
Since the default we want is the boolean value `True`, and we only need the parameters to disable those actions, we should use `action='store_false'` and rename the parameters to something like `--no-commit-updates` and `--no-push-updates`. We also need to pass `dest=commit_updates` and `dest=push_updates`.
If you really wanted to keep the parameter names, you could pass `type=ast.literal_eval`, so `--commit-updates False` would set the variable to the bolean `False`, but I don't think this is a good approach.

lib/config.py
alanoe: Agreed. Done

tools/upgrade_versions.py
alanoe: I prefer to do this split in a separate patch then. Done

tools/upgrade_versions.py
alanoe: It was not already following it before. I'll use the opportunity to fix it anyway.

tools/upgrade_versions.py

tools/upgrade_versions.py
alanoe: Done

lib/rpm_package.py

lib/config.py

tools/upgrade_versions.py

tools/upgrade_versions.py
ghost: This does not belong to this patch. It is duplicated.

tools/upgrade_versions.py
```diff
@@ -229,13 +229,18 @@ def run(CONF):
...
-    REQUIRED_PARAMETERS = ["push_repo_url", "push_repo_branch",
-                           "committer_name", "committer_email"]
+    updater_name = CONF.get('default').get('updater_name')
+    updater_email = CONF.get('default').get('updater_email')
+
+    if push_updates:
+        REQUIRED_PARAMETERS = ["push_repo_url", "push_repo_branch"]
+    else:
+        REQUIRED_PARAMETERS = []
+    REQUIRED_PARAMETERS += ["updater_name", "updater_email"]
```
ghost: You could set `REQUIRED_PARAMETERS = ["updater_name", "updater_email"]` first and remove the else statement.

ghost (changes requested):

tools/upgrade_versions.py

ghost (approved):

PR #175, opened 2017-02-03
Change runtime dependencies attrib name and behavior in build

alanoe:
- Rename dependencies attrib in yaml to runtime_dependencies
- Stop installing runtime dependencies in chroot used to build package
- Stop building runtime dependencies before dependent package

### DIFF
--- a/lib/rpm_package.py
+++ b/lib/rpm_package.py
@@ -161,9 +161,15 @@ class RPM_Package(Package):
             self.download_build_files = files.get('download_build_files', [])
 
             # list of dependencies
+            for dep_name in files.get('install_dependencies', []):
+                dep = RPM_Package.get_instance(dep_name, self.distro)
+                self.install_dependencies.append(dep)
+
+            # keeps backward compatibility with old yaml files which have 'dependencies'
+            # instead of 'install_dependencies'
             for dep_name in files.get('dependencies', []):
                 dep = RPM_Package.get_instance(dep_name, self.distro)
-                self.dependencies.append(dep)
+                self.install_dependencies.append(dep)
 
             for dep_name in files.get('build_dependencies', []):
                 dep = RPM_Package.get_instance(dep_name, self.distro)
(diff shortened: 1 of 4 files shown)


lib/rpm_package.py
```diff
@@ -161,6 +161,12 @@ def _load(self):
             self.download_build_files = files.get('download_build_files', [])
 
             # list of dependencies
+            for dep_name in files.get('runtime_dependencies', []):
+                dep = RPM_Package.get_instance(dep_name, self.distro)
+                self.dependencies.append(dep)
```
ghost: Since you're updating the yaml attribute, it would be better to rename the variable accordingly to avoid ambiguity, don't you think?
Like I mentioned in https://github.com/open-power-host-os/versions/pull/118#issuecomment-277278343, I prefer the name `install_dependencies`.

ghost (changes requested):

lib/scheduler.py
```diff
@@ -47,10 +47,12 @@ def _dfs(self, packages, visited):
...
                 visited.append(p)
-                if p.dependencies:
-                    order.extend(self._dfs(p.dependencies, visited))
                 if p.build_dependencies:
                     order.extend(self._dfs(p.build_dependencies, visited))
                 order.append(p)
+                # runtime dependencies do not need to be built before the package,
+                # they can be built anytime
+                if p.dependencies:
+                    order.extend(self._dfs(p.dependencies, visited))
```
ghost: If they can be built at any time, there's no need to deliberately place them after the dependent package.
One reason it could be useful to have the install dependency built before is that it could be installed as soon as the RPM is built, while the dependent package is still building. The opposite could not be done. 
Also, if the dependency build fails, we'd end with a dependent RPM that could not be installed, while, with the opposite, if any package build fails, all the ones that succeeded are installable.

ghost (changes requested):

lib/rpm_package.py

ghost (approved):

PR #177, opened 2017-02-03
Make rebuilding packages smarter

ghost:
This changes the default behavior of build-packages command to only build packages when needed.
Any of the following criteria is enough to rebuild a package:
- It has no previously cached build results
- Any of Its source files (yaml, spec, etc...) were modified
- It has build dependencies that were rebuilt
- The option --force-rebuild was used (forces rebuild of all packages at once)

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -21,6 +21,7 @@ default:
    --plugin-option=tmpfs:required_ram_mb=39800"
  commit_updates: True
  push_updates: True
+ force_rebuild: False
 iso:
     iso_name: "OpenPOWER-Host_OS"
     log_file: "/var/log/host-os/iso.log"
--- a/lib/build_manager.py
+++ b/lib/build_manager.py
@@ -15,11 +15,13 @@
 
 import logging
 
+from lib import config
 from lib import exception
 from lib.packages_manager import PackagesManager
 from lib.rpm_package import RPM_Package
 from lib.scheduler import Scheduler
 
+CONF = config.get_config().CONF
 LOG = logging.getLogger(__name__)
 
 
@@ -30,10 +32,11 @@ class BuildManager(object):
         self.repositories = None
 
     def __call__(self):
+        force_rebuild = CONF.get('default').get('force_rebuild')
         try:
             self.packages_manager.prepare_packages(
                 packages_class=RPM_Package, distro=self.distro,
-                download_source_code=False)
+                download_source_code=False, force_rebuild=force_rebuild)
         # distro related issues
         except (exception.DistributionNotSupportedError,
                 exception.DistributionVersionNotSupportedError,
--- a/lib/build_manager.py
+++ b/lib/build_manager.py
@@ -15,11 +15,10 @@
 
 import logging
 
-import lib.centos
 from lib import exception
-import lib.scheduler
 from lib.packages_manager import PackagesManager
 from lib.rpm_package import RPM_Package
+from lib.scheduler import Scheduler
 
 LOG = logging.getLogger(__name__)
 
@@ -46,5 +45,5 @@ class BuildManager(object):
         self.build()
 
     def build(self):
-        scheduler = lib.scheduler.Scheduler()
+        scheduler = Scheduler()
         self.distro.build_packages(scheduler(self.packages_manager.packages))
--- a/lib/mockbuilder.py
+++ b/lib/mockbuilder.py
@@ -17,6 +17,7 @@ from functools import partial
 
 import datetime
 import logging
+import glob
 import os
 import shutil
 
@@ -36,7 +37,6 @@ class Mock(build_system.PackageBuilder):
         super(Mock, self).__init__()
         binary_file = CONF.get('default').get('mock_binary')
         extra_args = CONF.get('default').get('mock_args')
-        self.result_dir = CONF.get('default').get('result_dir')
         self.build_dir = None
         self.archive = None
         self.timestamp = datetime.datetime.now().isoformat()
@@ -129,7 +129,9 @@ class Mock(build_system.PackageBuilder):
             cmd = self.common_mock_args
             install = " --install"
             for dep in package.build_dependencies:
-                install = " ".join([install, " ".join(dep.result_packages)])
+                dep_rpms_glob = os.path.join(dep.build_results_dir, "*.rpm")
+                rpm_files = glob.iglob(dep_rpms_glob)
+                install = " ".join([install, " ".join(rpm_files)])
 
             cmd = cmd + install
             LOG.info("%s: Installing dependencies on chroot" % package.name)
@@ -145,18 +147,18 @@ class Mock(build_system.PackageBuilder):
         shutil.rmtree(self.build_dir)
 
     def _save_rpm(self, package):
-        if not os.path.exists(self.result_dir):
-            LOG.info("Creating directory to store RPM at %s " %
-                     self.result_dir)
-            os.makedirs(self.result_dir)
-            os.chmod(self.result_dir, 0777)
+        if not os.path.exists(package.build_results_dir):
+            LOG.info("Creating directory to store RPMs at %s " %
+                     package.build_results_dir)
+            os.makedirs(package.build_results_dir)
+            os.chmod(package.build_results_dir, 0777)
 
-        LOG.info("%s: Saving RPMs at %s" % (package.name, self.result_dir))
+        LOG.info("%s: Saving RPMs at %s"
+                 % (package.name, package.build_results_dir))
         for f in os.listdir(self.build_dir):
             if f.endswith(".rpm") and not f.endswith(".src.rpm"):
-                LOG.info("Saving %s at result directory %s" % (f,
-                         self.result_dir))
+                LOG.info("Saving %s at result directory %s" % (
+                    f, package.build_results_dir))
                 orig = os.path.join(self.build_dir, f)
-                dest = os.path.join(self.result_dir, f)
+                dest = os.path.join(package.build_results_dir, f)
                 shutil.move(orig, dest)
-                package.result_packages.append(dest)
--- a/lib/mockbuilder.py
+++ b/lib/mockbuilder.py
@@ -77,7 +77,7 @@ class Mock(build_system.PackageBuilder):
             raise
 
         msg = "%s: Success! RPMs built!" % (package.name)
-        self._save_rpm(package)
+        self._copy_rpms(self.build_dir, package.build_results_dir)
         LOG.info(msg)
         if not CONF.get('default').get('keep_builddir'):
             self._destroy_build_directory()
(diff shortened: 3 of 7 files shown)


olavphibm:
The first two commits belong to https://github.com/open-power-host-os/builds/pull/175

lib/package.py
```diff
@@ -56,7 +56,8 @@ def __init__(self, name):
         self.download_source = None
         self.install_dependencies = []
         self.build_dependencies = []
-        self.result_packages = []
+        self.artifacts_dir = os.path.join(
```
alanoe: pls be consistent in the terminology, use either "artifacts" or "results". I prefer the latter, as it is a more universally know term. "artifacts" is Jenkins terminology.

config.yaml
```diff
@@ -21,6 +21,7 @@ default:
    --plugin-option=tmpfs:required_ram_mb=39800"
  commit_updates: True
  push_updates: True
+ build_dependencies: True
```
alanoe: I don't think that this global attribute is needed, I'd prefer to have a clean_build_dependencies_cache boolean option instead. If this option is disabled, the dependencies reuse would be dynamic instead of the static approach you suggest here. We would not assume that all dependencies are already pre-built or not; for each package, the build system auto-detects if the dependencies of a package are already built in the system and reuses them if they are. To achieve this, we just need to choose an adequate primary key for a package build (eg `<package name`>-`<commit ID`>) and somehow keep an updated table of the built packages. One way to achieve this is by saving each build package in a result dir of format <primary_key>, later parsing this directory structure to infer if the dependency package was already build or not and reuse the package if it was already built.

lib/build_manager.py
```diff
@@ -46,5 +48,11 @@ def __call__(self):
...
     def build(self):
-        scheduler = lib.scheduler.Scheduler()
-        self.distro.build_packages(scheduler(self.packages_manager.packages))
+        packages_order = self.packages_manager.packages
+        if CONF.get("default").get("build_dependencies"):
+            scheduler = lib.scheduler.Scheduler()
+            packages_order = scheduler(packages_order)
+        else:
+            LOG.info("Not scheduling build of dependencies")
+            LOG.info("Building packages: {}".format(packages_order))
```
alanoe: I don't agree with this. The build scheduler should always be used, there is no reason to let user specify the build order.

lib/mockbuilder.py
```diff
@@ -17,6 +17,7 @@
```
alanoe, line 17: It seems to me that, with this approach, if you desire to build gcc and 2 packages which depend on gcc, gcc will be built 3 times. This does not make sense to me. Maybe I misunderstood? 

I don't think we need to keep the built build dependencies in a package-specific directory, I'd prefer to have a directory for each package built (named after a primary key) and keep them all under a single directory, as mentioned in the other patch of this PR.

alanoe (changes requested):

config.yaml
ghost: I don't think using the commit ID as a primary key is a good idea. There are many other factors that may require rebuilding a package. Altering the spec file or modifying the build scripts, for example. I agree with checking if the dependencies are already built by checking if a directory `<package_name>` exists, but I think the default should be to always rebuild the dependencies and the decision not to do this should be mostly manual.

lib/build_manager.py
ghost: Flexibility is always nice, but yeah, I don't see any good reason to do this either.

lib/mockbuilder.py
ghost: The way I've done it, only the packages explicitly passed would be built, so passing `gcc, kernel` would only build gcc once. I'm applying your other suggestions though, so this will be redone.

I don't get your second point, doesn't your suggestion of using a primary key lead to package (and commit) specific directories?

lib/package.py
ghost: The class' name is `Package`, so `Package.results` does not make much sense to me. I admit `Package.artifacts` doesn't either. What do you think of `Package.build_results_dir`?

lib/mockbuilder.py
alanoe: OH> I don't get your second point, doesn't your suggestion of using a primary key lead to package (and commit) specific directories?

I was under the impression in my first quick read of the code that, if you are building A and B and C are dependencies, you'd put B and C rpms inside A dir. In that case, if C was a dependency for D, you would not find C inside D dir and rebuild C.

config.yaml
alanoe: My main point here is automating the act of inferring if *each* package must be rebuilt instead of having a global manual option to do it, the commit ID was just the first idea that came in my mind. Better ideas are welcome. =)   Another (better) way is to keep the date of the last modification in the package directory and rebuild if there were any modifications since then. This would cover adding of files to SOURCES, changes in the spec and changes in the yaml. The only case this approach would not cover is a modification to the builds git repo, but then you could explicitly use the option to clean the cache I've mentioned before.

lib/package.py
alanoe: +1

lib/rpm_package.py
```diff
@@ -187,19 +187,25 @@ def _load(self):
                 self.build_files = None
             self.download_build_files = files.get('download_build_files', [])
 
+
+            rebuild_dependencies = CONF.get('default').get('force_rebuild')
```
farosas: Why can't you use `self.rebuild` here?

lib/packages_manager.py
```diff
@@ -25,7 +25,14 @@
...
     def __init__(self, packages_names):
-        self.packages_names = packages_names
+        """
+        Class that manages actions on packages.
+
+        Args:
+            packages_names: names of the packages that will be processed.
+                Leave empty to discover all available packages.
+        """
+        self.packages_names = packages_names or config.discover_packages()
```
farosas: I don't get how moving the package discovery to here accomplishes what you describe in the commit message. Am I missing something here?

lib/rpm_package.py
ghost: This option had a slightly different meaning in a previous version of the PR.
Changing to `self.rebuild`

lib/packages_manager.py
ghost: This was needed for a previous version of the PR that had a slightly different functionality. I am dropping this patch.

lib/mockbuilder.py
```diff
@@ -17,6 +17,7 @@
 
 import datetime
```
alanoe: in the commit longer msg, form => from

lib/mockbuilder.py
```diff
@@ -128,9 +128,7 @@ def _install_external_dependencies(self, package):
             cmd = self.common_mock_args
             install = " --install"
             for dep in package.build_dependencies:
-                dep_rpms_glob = os.path.join(dep.build_results_dir, "*.rpm")
-                rpm_files = glob.iglob(dep_rpms_glob)
-                install = " ".join([install, " ".join(rpm_files)])
+                install = " ".join([install, " ".join(dep.cached_build_results)])
```
alanoe: I'd prefer "cached_built_rpm_files" to make clearer this variable is a list of files paths.

lib/package.py
```diff
@@ -56,6 +59,7 @@ def __init__(self, name):
         self.download_source = None
         self.install_dependencies = []
         self.build_dependencies = []
+        self.build_cache_dir = os.path.join(BUILD_CACHE_DIR, self.name)
```
alanoe: Not sure if we need to keep build_cache_dir in Package if it easily inferred from cache dir and package name, but up to you. The same applies to build_results_dir BTW

lib/mockbuilder.py
```diff
@@ -145,18 +145,23 @@ def _create_build_directory(self, package):
...
+        """
+        if not os.path.exists(target_dir):
+            LOG.debug("Creating directory to store RPMs at %s " % target_dir)
+            os.makedirs(target_dir)
+
+        LOG.info("Copying RPMs from %s to %s" % (source_dir, target_dir))
+        for rpm_file in os.listdir(source_dir):
+            if rpm_file.endswith(".rpm") and not rpm_file.endswith(".src.rpm"):
+                LOG.info("Copying RPM file: %s" % rpm_file)
+                orig = os.path.join(source_dir, rpm_file)
```
alanoe: - I'd prefer to be consistent with source_dir and target_dir in these orig/dest variables names
- Also, I'd prefer to have some consistency in the names of variables which refer to the file name and to the file path. You are using rpm_file for file name and orig/dest. Either use rpm_file_name and rpm_file_path, source_file_name and source_file_path or something similar.

lib/mockbuilder.py
```diff
@@ -145,18 +145,23 @@ def _create_build_directory(self, package):
...
+                dest = os.path.join(target_dir, rpm_file)
+                shutil.copy(orig, dest)
```
alanoe: It is not clear to me why you are copying instead of moving. Could you explain better? 

Minor: I'd prefer to move this functional change to a separate patch and leave only refactoring here.

config.yaml
```diff
@@ -21,6 +21,7 @@ default:
    --plugin-option=tmpfs:required_ram_mb=39800"
```
alanoe, line 21: minor: In commit longer msg, imho "This patch adds an option" is clearer than "There's an option"

alanoe (changes requested):
Minor changes only

farosas (approved):

lib/mockbuilder.py
ghost: Done

config.yaml
ghost: Done

lib/mockbuilder.py
ghost: Done

lib/mockbuilder.py
ghost: There are two benefits: allowing the function to be reusable when copying from the cache to the results dir and keeping the build artifacts in the builds directory to be more easily found. Explaining in a separate (previous) commit.

lib/package.py
ghost: @gutoyr will work on changing the cache and results paths. Let's see if it's still needed after his changes.

lib/mockbuilder.py
ghost: This is a property of the `Package` class, so we can't have RPM in its name.

alanoe: The point is that "results" are a generic term, it could be logs, built packages or extra outputs (eg a yum repository config file). 

You may use cached_built_files or cached_built_files_paths.

ghost: I've taken some time to rewrite all occurrences of "results" to "built_files" and I don't think its an improvement. It makes logs and docs harder to read. "Results", although being generic, is a simpler term.
For example: "Build results found" vs. "Previously built files found".
It is not immediately clear that "built files" resulted from a previous build, so we'd have to take extra care to explain this every time.
I'm keeping this as is.

PR #178, opened 2017-02-04
Fix access of proxy arguments

ghost:
The dictionary containing the arguments was not being accessed
correctly, so the proxy options would always be disabled.


alanoe (approved):
More one proof that this feature was never used...

ghost (approved):

PR #179, opened 2017-02-07
Add proxy config to SVN checkout

ghost:
Is this proxy support still a requirement?

I wish we could drop this from our code, as it appears no one is in need of this _feature_

### DIFF
--- a/lib/repository.py
+++ b/lib/repository.py
@@ -201,9 +201,28 @@ class SvnRepository():
         """
         LOG.info("Checking out repository from '%s' into '%s'" %
                  (remote_repo_url, repo_path))
+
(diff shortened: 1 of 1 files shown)


olavphibm:
start build

alanoe:
LT> Is this proxy support still a requirement?

Nobody is using this option, but we were asked to add it so a IBM team in Australia could more easily use the OpenPower Host OS build scripts. Unfortunately, while there is the possibility that maybe they will use the build scripts, I doubt we'll convince leadership to remove the option.

alanoe (approved):

ghost (approved):

PR #180, opened 2017-02-07
Add distro options to host.os.py's command line parser

alanoe:


lib/config.py
```diff
@@ -50,6 +50,12 @@
         dict(help='Keep build directory and its logs and artifacts.',
              action='store_true'),
 }
+DISTRO_ARGS = {
+    ('--distro-name',):
+        dict(help='Base Linux distribution', default=''),
```
ghost: Why have this default? The default value `None` is just as good, and it will be replaced by the value in `config.yaml` anyway.

ghost (approved):

farosas (approved):

ghost:
@alanoe Mentioned that we're probably not gonna need https://github.com/open-power-host-os/builds/pull/180/commits/91e6affefa1c89a1daea92defee20c76f4b5a069. Let's put this PR on hold until we have the Jenkins jobs for validating previous versions and make sure they work with this code.

alanoe:
start build

PR #181, opened 2017-02-09
Add script to update versions repo README

ghost:
Regenerates the table of supported software-versions.

### DIFF
--- a/host_os.py
+++ b/host_os.py
@@ -23,6 +23,7 @@ from tools import build_iso
 from tools import build_package
 from tools import create_release_notes
 from tools import setup_environment
+from tools import update_versions_readme
 from tools import upgrade_versions
 
 INSUFFICIENT_PRIVILEGE_ERROR = 3
@@ -34,6 +35,7 @@ SUBCOMMANDS = {
     'build-package': build_package,
     'release-notes': create_release_notes,
     'upgrade-versions': upgrade_versions,
+    'update-versions-readme': update_versions_readme,
     'set-env': setup_environment,
     'build-iso': build_iso,
 }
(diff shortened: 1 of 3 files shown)


host_os.py
```diff
@@ -34,6 +35,7 @@
     'build-package': build_package,
     'release-notes': create_release_notes,
     'upgrade-versions': upgrade_versions,
+    'update-versions-readme': update_versions_readme,
```
alanoe, line 38: I'd prefer to do this in upgrade-versions instead of creating a new subcommand. Both will always be executed together.

alanoe (changes requested):
See my comment

host_os.py
ghost: I don't agree this is a good option.

Keeping it a separate command makes it more flexible, in case we only want to update the README. For example, if we want to update the README when a package (not covered by `upgrade-versions`) is changed? That wouldn't work.

Refer to infrastructure/pull/28

ghost:
If I understood this correctly, whenever someone opens a PR in versions repo (including the ones created by weekly builds), a job will start in Jenkins to update the README.md file and it will create another commit in the bot's repo with the updated README file.
Every PR that updates a package version will then create another PR with the correct versions in the README file. The person that is taking care of the weekly builds will then have to check every time if any version has changed to open the second PR when needed.
If that's correct, I don't like this approach.
I believe the same commit that modifies the versions in the spec files should also modify the README. Creating a second commit for the README would make the first commit inconsistent.
One solution I see is calling the function that updates the README file into the upgrade-version command. I agree that executing the command standalone can be useful for when versions are upgraded manually (or new packages are added). So we could also add this subcommand, calling the function as it's done in upgrade-versions.

alanoe:
+1 with Olav, it is bad to have 2 commits for an atomic operation. Either we:

1) automatically update a commit when it is pushed to include the update in the README

2) add a new subcommand which calls a function which is automatically called on upgrade-versions. This subcommand must be called manually if someone manually edits a version in versions repository.

(2) seems simpler to implement and it is closer to what you have already done.

Murilo said Lucas is on vacation. May I work on this and resubmit the PR?

ghost:
@alanoe I agree in implementing (2). Go ahead if you want to take this.

alanoe:
I've implemented the change mentioned above. I'll close this pull request and open a new one.

PR #185, opened 2017-02-14
Modify github.io pages creation to use tags from versions

lisibann:
The versions repo will have tags for each weekly build and tags for major releases. The github.io page needs to be altered to include the tag string in the pages titles.

### DIFF
--- a/lib/versions_repository.py
+++ b/lib/versions_repository.py
@@ -111,3 +111,23 @@ def update_versions_in_readme(versions_repo, distro, packages_names):
     output_readme_path = os.path.join(versions_repo.working_tree_dir,
                                       'README.md')
     replace_html_table(output_readme_path, html_table)
+
+
+def read_version_and_milestone(versions_repo):
+    """
+    Read current version and milestone (alpha or beta) from VERSION file
+
+    Args:
+        versions_repo (GitRepository): packages metadata git repository
+    Returns:
+        version_milestone (str): version and milestone. Format:
+            <version>-<milestone>, valid milestone values: alpha, beta
+    """
+    version_file_path = os.path.join(versions_repo.working_tree_dir, 'VERSION')
+    version_milestone = ""
+    with open(version_file_path, 'r') as version_file:
+        #ignore first line with file format information
+        version_file.readline()
+        version_milestone = version_file.readline().strip('\n')
+
+    return version_milestone
--- a/tools/create_release_notes.py
+++ b/tools/create_release_notes.py
@@ -26,6 +26,7 @@ from lib import packages_manager
 from lib import repository
 from lib import rpm_package
 from lib.versions_repository import setup_versions_repository
+from lib.versions_repository import read_version_and_milestone
 
 CONF = config.get_config().CONF
 LOG = logging.getLogger(__name__)
@@ -34,6 +35,7 @@ RELEASE_FILE_NAME_TEMPLATE = "{date}-release.markdown"
 RELEASE_FILE_CONTENT_TEMPLATE = """\
 ---
 title: OpenPOWER Host OS release
+release_tag: {release_tag}
 layout: release
 builds_commit: {builds_commit}
 versions_commit: {versions_commit}
@@ -60,7 +62,7 @@ class PackageReleaseInfo(object):
             ("name", self.package.name),
             ("clone_url", main_git_source["src"] if main_git_source else ""),
             ("version", self.package.version),
-            ("release",  self.package.release),
+            ("release", self.package.release),
             ("branch", main_git_source["branch"] if main_git_source else ""),
             ("commit_id",
              main_git_source["commit_id"] if main_git_source else ""),
@@ -70,14 +72,14 @@ class PackageReleaseInfo(object):
         return string
 
 
-def write_version_info(release, file_path, versions_repo, packages):
+def write_version_info(release_tag, file_path, versions_repo, packages):
     """
     Write release information to a file.
     It contains packages names, branches and commit IDs.
     """
-    LOG.info("Creating release {release} information".format(**locals()))
-    format_dict = {"release": release}
+    LOG.info("Creating release {release_tag} information".format(**locals()))
 
+    format_dict = {"release_tag": release_tag}
     format_dict["builds_commit"] = (
         repository.GitRepository(".").head.commit.hexsha)
     format_dict["versions_commit"] = versions_repo.head.commit.hexsha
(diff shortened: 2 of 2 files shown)


lib/versions_repository.py
```diff
@@ -25,3 +25,17 @@ def setup_versions_repository(config):
         raise
 
     return versions_repo
+
+def read_version_and_milestone(versions_repo):
+    """
+    Read current version and milestone (alpha or beta) from VERSION file
+    """
```
alanoe: pls document versions_repo type

lib/versions_repository.py
```diff
@@ -25,3 +25,17 @@ def setup_versions_repository(config):
...
+    version_path =  os.path.join(versions_repo.name, 'VERSION')
+    version_milestone_format = ""
+    version_milestone = ""
+    with open(version_path, 'r') as file_version:
+        version_milestone_format = file_version.readline()
```
alanoe: you could use _ instead of version_milestone_format or add a ignore directive to avoid a Pylint unused variable warning

tools/create_release_notes.py
```diff
@@ -34,6 +35,7 @@
 RELEASE_FILE_CONTENT_TEMPLATE = """\
 ---
 title: OpenPOWER Host OS release
+release_tag: {release_tag}
```
alanoe, line 38: imho we do not need the "tag" word in all variables names here. The fact that the release is a git repository tag is irrelevant here, we should avoid unnecessary coupling.

alanoe (changes requested):

lib/versions_repository.py
lisibann: done

lib/versions_repository.py
lisibann: done

tools/create_release_notes.py
lisibann: I'll keep "tag". It should match the git tag to be created for versions git repository.

lib/versions_repository.py
```diff
@@ -25,3 +25,19 @@ def setup_versions_repository(config):
...
+
+    Args:
+        versions_repo (GitRepository): packages metadata git repository
+    """
+    version_path = os.path.join(versions_repo.name, 'VERSION')
+    version_milestone = ""
+    with open(version_path, 'r') as file_version:
+        _ = file_version.readline()
+        version_milestone = file_version.readline().strip('\n')
+    file_version.close()
```
farosas: No need to close the file here, the context manager does it automatically.

farosas (changes requested):

lib/versions_repository.py
lisibann: done

lib/versions_repository.py
```diff
@@ -25,3 +25,18 @@ def setup_versions_repository(config):
...
     return versions_repo
+
+def read_version_and_milestone(versions_repo):
+    """
+    Read current version and milestone (alpha or beta) from VERSION file
+
+    Args:
+        versions_repo (GitRepository): packages metadata git repository
+    """
+    version_path = os.path.join(versions_repo.name, 'VERSION')
```
ghost: Use `versions_repo.working_tree_dir` instead of name. This will make the function executable from any directory.
I'd rename the variable to `version_file_path`.

lib/versions_repository.py
```diff
@@ -25,3 +25,18 @@ def setup_versions_repository(config):
...
+    version_milestone = ""
+    with open(version_path, 'r') as file_version:
```
ghost: `file_version` looks like we're getting the version of the file. Since it's a file containing the version of the repo, use `version_file` instead.

lib/versions_repository.py
```diff
@@ -25,3 +25,18 @@ def setup_versions_repository(config):
...
+        _ = file_version.readline()
```
ghost: You dont have to set the function result to a variable. Just write  `file_version.readline()`.

lib/versions_repository.py
```diff
@@ -25,3 +25,18 @@ def setup_versions_repository(config):
...
+
+def read_version_and_milestone(versions_repo):
+    """
+    Read current version and milestone (alpha or beta) from VERSION file
+
+    Args:
+        versions_repo (GitRepository): packages metadata git repository
+    """
+    version_path = os.path.join(versions_repo.name, 'VERSION')
+    version_milestone = ""
```
ghost: There's no need to initialize this variable.

lib/versions_repository.py
```diff
@@ -25,3 +25,18 @@ def setup_versions_repository(config):
         raise
 
     return versions_repo
+
+def read_version_and_milestone(versions_repo):
+    """
+    Read current version and milestone (alpha or beta) from VERSION file
+
+    Args:
+        versions_repo (GitRepository): packages metadata git repository
```
ghost, line 121: Since youre documenting the args, document the results too.

lib/versions_repository.py
```diff
@@ -25,3 +25,18 @@ def setup_versions_repository(config):
...
+    """
+    version_path = os.path.join(versions_repo.name, 'VERSION')
+    version_milestone = ""
+    with open(version_path, 'r') as file_version:
+        _ = file_version.readline()
+        version_milestone = file_version.readline().strip('\n')
```
ghost: You could return right away here instead of using a variable.

ghost (changes requested):

lib/versions_repository.py
lisibann: done

lib/versions_repository.py
lisibann: done

lib/versions_repository.py
lisibann: done

lib/versions_repository.py
lisibann: done

lib/versions_repository.py
lisibann: done

ghost (approved):
You'll have to rebase this on top of master.

lib/versions_repository.py
```diff
@@ -25,3 +25,21 @@ def setup_versions_repository(config):
...
     return versions_repo
+
+def read_version_and_milestone(versions_repo):
+    """
+    Read current version and milestone (alpha or beta) from VERSION file
+
+    Args:
+        versions_repo (GitRepository): packages metadata git repository
+    Returns:
+        version_milestone (str): version and milestone
```
alanoe: Pls document the expected format of this string (`<version`>-`<milestone`>, valid milestone values: alpha, beta)

alanoe: minor: version_and_milestone seems clearer to me than version_milestone, but up to you

alanoe (changes requested):
only minor stuff

lib/versions_repository.py
lisibann: done

alanoe (approved):

farosas (approved):

PR #187, opened 2017-02-15
Add documentation about how to build old versions

alanoe:


README.rst
```diff
@@ -94,6 +94,19 @@ or
...
+OpenPOWER Host OS 1.0 and 1.5 used an older CentOS (7.2) as base version. 
+To build it, you must do the following changes in this git repository
+before building:
+ * Replace "http://mirror.centos.org/altarch/7" with "http://vault.centos.org/altarch/7.2.1511/"
+   in mock_configs/CentOS/7/CentOS-7-ppc64le.cfg
+ * Replace distro_version value in config.yaml from "7" to "7.2"
+ * Replace "mock_config/CentOS/7" child subnode in config.yaml from "7" to "7.2"
+ * Replace build_version in config.yaml to v1.5.0 or v1.0.0
+
+These steps are documented in jenkins_jobs/build_old_host_os_versions/script.sh in the
```
ghost: You could place a direct link to the file here.

alanoe: Done

README.rst
```diff
@@ -94,6 +94,19 @@ or
...
+* Build old versions
+
+OpenPOWER Host OS 1.0 and 1.5 used an older CentOS (7.2) as base version. 
+To build it, you must do the following changes in this git repository
+before building:
+ * Replace "http://mirror.centos.org/altarch/7" with "http://vault.centos.org/altarch/7.2.1511/"
+   in mock_configs/CentOS/7/CentOS-7-ppc64le.cfg
+ * Replace distro_version value in config.yaml from "7" to "7.2"
+ * Replace "mock_config/CentOS/7" child subnode in config.yaml from "7" to "7.2"
+ * Replace build_version in config.yaml to v1.5.0 or v1.0.0
```
farosas: I think we should improve the presentation of this info. Some points:

1- the nested bullets are ugly in GitHub 
2- there is too much text. E.g. in first bullet, the reader needs to visually parse both mirror links before getting to the name of the file that needs to be edited
3- too many non-alphabetical characters make it harder to read

I suggest a more concise text, without nesting and providing commands to copy/paste:

---
* Build old versions

OpenPOWER Host OS 1.0 and 1.5 used an older CentOS (7.2) as base version. 

Change yum mirror to CentOS 7.2
```
$ sed -i 's!http://mirror.centos.org/altarch/7!http://vault.centos.org/altarch/7.2.1511/!' mock_configs/CentOS/7/CentOS-7-ppc64le.cfg
```

Change the distro version to 7.2
```
$ sed -i "s/'7'/'7.2'/g" config.yaml 
```

Inform which version to build
```
$ sed -i "s/build_version:.*/build_version: 'v1.5.0'/" config.yaml # use v1.5.0 or v1.0.0
```

---
*note: the seds might be improved a bit*
*note 2: I realize the second command hides two actions in one and that we have to be careful with replacing '7', but I think it's worth it.*

farosas (changes requested):

README.rst
alanoe: OK for me. I just prefer to use 2 more explicit seds then the one you use to update the distro version.

README.rst
```diff
@@ -94,6 +94,31 @@ or
 
 $ python host_os.py build-package --help
 
+* Build old versions
+
+OpenPOWER Host OS 1.0 and 1.5 used an older CentOS (7.2) as base version.
+To build one of them, do the following changes in this git repository
+before building:
+
+Update yum repository to CentOS 7.2
```
farosas: missing :: in all entries

farosas (changes requested):

README.rst
alanoe: Done

PR #188, opened 2017-02-15
Add option to update README in versions git repository to host_os.py

alanoe:
This option regenerates the table of supported packages versions.
Also, this patch automatically does this update then upgrade-versions
option is used.

### DIFF
--- a/host_os.py
+++ b/host_os.py
@@ -24,6 +24,7 @@ from tools import build_iso
 from tools import build_package
 from tools import create_release_notes
 from tools import setup_environment
+from tools import update_versions_in_readme
 from tools import upgrade_versions
 
 INSUFFICIENT_PRIVILEGE_ERROR = 3
@@ -35,6 +36,7 @@ SUBCOMMANDS = {
     'build-package': build_package,
     'release-notes': create_release_notes,
     'upgrade-versions': upgrade_versions,
+    'update-versions-readme': update_versions_in_readme,
     'set-env': setup_environment,
     'build-iso': build_iso,
 }
--- a/lib/versions_repository.py
+++ b/lib/versions_repository.py
@@ -1,8 +1,12 @@
 import logging
 import os
 
+from lxml import etree
+
 from lib import exception
+from lib import packages_manager
 from lib import repository
+from lib import rpm_package
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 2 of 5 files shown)


tools/upgrade_versions.py
```diff
@@ -28,6 +28,7 @@
 from lib import rpm_package
 from lib.utils import replace_str_in_file
 from lib.versions_repository import setup_versions_repository
+from tools.update_versions_readme import update_versions_in_readme
```
ghost: I believe reusable functions should be placed in `lib`, so we wouldn't have any `import tools`

host_os.py
```diff
@@ -34,6 +35,7 @@
     'build-package': build_package,
     'release-notes': create_release_notes,
     'upgrade-versions': upgrade_versions,
+    'update-versions-readme': update_versions_readme,
```
ghost: This name is inconsistent with "upgrade-versions" above. I prefer to use "update" for everything, but that would require changing the infrastructure repo too. We can leave this change for later if you feel it's too much trouble for now.

tools/update_versions_readme.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def update_versions_in_readme(versions_repo, distro, packages):
+    """
+    Update packages versions in README
+
+    Args:
+        versions_repo (GitRepository): versions git repository handler
+        distro (distro.LinuxDistribution): Linux distribution
+        packages ([str]): list of packages which versions must be updated
```
ghost: "whose versions"

ghost: I'd name this variable `package_names` to avoid confusing with the `Package` type.

tools/update_versions_readme.py
```diff
@@ -0,0 +1,165 @@
...
+    LOG.info("Generating packages versions HTML table from packages: %s",
+             ", ".join(packages))
+    pm = packages_manager.PackagesManager(packages)
+    # TODO: this is coupled with RPM-based Linux distributions
+    pm.prepare_packages(packages_class=rpm_package.RPM_Package,
+                        download_source_code=False, distro=distro)
+
+    html_table = generate_html_table(pm.packages)
+    readme_template_path = os.path.join(versions_repo.working_tree_dir,
+                                        'README.md.in')
```
ghost: Like I said to you, I prefer not to have this file and modify the README.md directly.

ghost (changes requested):

tools/upgrade_versions.py
alanoe: I was just trying to alter Lucas' patch as less as possible. Done.

TBH I do not understand why we have this tools directory, IMHO entire code would be in lib and executable scripts would simply call them, as we had in PowerKVM project.

host_os.py
alanoe: I'll send a patch in another PR which renames upgrade-versions/upgrade_versions to update-versions/update_versions. Pls let us keep this inconsistency for now.

tools/update_versions_readme.py
alanoe: Done

alanoe: Done

tools/update_versions_readme.py
alanoe: As we discussed, this will make the versions table replacement a little harder, but no problem. Working on it

lib/versions_repository.py
```diff
@@ -25,3 +29,89 @@ def setup_versions_repository(config):
...
+
+    with file(file_path, "r") as f:
+        lines = f.readlines()
+
+    new_lines = []
+    in_table = False
+    for line in lines:
+        if not in_table and "<table>" in line:
+            in_table = True
+            new_lines.append(HTML_TABLE_TEMPLATE_VARIABLE)
```
ghost: Do you really need this constant?
You could append the `html_table` directly here, and continue to ignore the table lines.

tools/upgrade_versions.py
```diff
@@ -254,6 +255,8 @@ def run(CONF):
         pkg_version.update(updater_name, updater_email)
         pkg.unlock()
 
+    update_versions_in_readme(versions_repo, distro, packages_to_update)
```
ghost: This will create a table with only the packages being updated. We'll want a complete table instead, so you could either pass `config.discover_packages()` or discover the packages inside the function and disallow generating incomplete tables.

ghost (changes requested):

lib/versions_repository.py
alanoe: That also works. Done

tools/upgrade_versions.py

lib/versions_repository.py
```diff
@@ -1,8 +1,12 @@
 import logging
 import os
 
+from lxml import etree
```
farosas, line 4: Need to update dependencies file

farosas (changes requested):

lib/versions_repository.py
alanoe: python-lxml is already in rpm_requirements.txt

ghost (approved):

farosas (approved):

PR #190, opened 2017-02-17
Refactor config

alanoe:
- Make options names consistent
- Sync options in config file and command line


lib/config.py
```diff
@@ -269,9 +270,11 @@ def parse(self):
 
         # update iso node with iso subcommand args and then drop them from args
         for key, value in args.items():
-            if key in config['iso']:
-                config['iso'][key] = value
-                args.pop(key)
+            for node in COMMAND_TO_CONFIG_NODE.values():
+                if key in config[node]:
```
ghost: This will never hold for `release-notes` because the key contains the "root node" name, e.g. `release-notes.repo-url`, while `config[node]` contains only the suffix.

lib/config.py
```diff
@@ -67,9 +67,9 @@
              default=''),
 }
 RELEASE_NOTES_ARGS = {
-    ('--release-notes-repo-url',):
+    ('--release-notes.repo-url',):
```
ghost: I don't know if you're doing it in a later patch, but since you're adding the dot, you may as well replace the hyphens with underscores to make it uniform with the yaml file.

lib/package.py
```diff
@@ -138,7 +138,7 @@ def _download_source_code(self):
                  (self.name, self.clone_url))
         self._setup_repository(
             dest=CONF.get('default').get('repositories_path'),
-            branch=CONF.get('default').get('branch'))
+            branch='powerkvm-v3.1.1')
```
ghost: This is not necessary. The branch is only used for the upgrade-versions script, for which backwards compatibility is not needed. It has also been changed to `hostos-devel` in all packages, so this is outdated.
I'd change this to `None`.

config.yaml
```diff
@@ -1,6 +1,6 @@
 default:
  log_file: "/var/log/host-os/builds.log"
- repositories_path: "/var/lib/host-os/repositories"
+ packages_repos_target_path: "/var/lib/host-os/repositories"
```
ghost: Just a heads up: this will likely conflict with the patches @gutoyr is working on.
I also believe the versions and open-power-host-os repositories will be kept in the same directory, so having packages in the name will not make sense. I may be wrong.

config.yaml
```diff
@@ -1,8 +1,8 @@
 default:
  log_file: "/var/log/host-os/builds.log"
  packages_repos_target_path: "/var/lib/host-os/repositories"
- build_versions_repository_url: "https://github.com/open-power-host-os/versions.git"
- build_version: 'master'
+ packages_metadata_repo_url: "https://github.com/open-power-host-os/versions.git"
```
ghost: Since this is exposed to the user and a common option to use, I'd prefer to use a more relatable name, like `packages_versions_repo_url`.

config.yaml
```diff
@@ -17,21 +27,16 @@ common:
...
-build_packages:
- packages:
- packages_repos_target_path: "/var/lib/host-os/repositories"
- keep_build_dir: False
- result_dir:  "./result"
- force_rebuild: False
 build_release_notes:
  repo_url: "https://github.com/open-power-host-os/open-power-host-os.github.io.git"
  repo_branch: "master"
+ push_repo_url: "git@github.com:open-power-host-os/open-power-host-os.github.io.git"
```
ghost: Do not set this default value. We shouldn't push directly to the main repository (it may happen by accident) and any other user will not have permission to do so. We should make it explicit that this parameter is required and needs to be set up properly.

config.yaml
```diff
@@ -80,3 +85,6 @@ build_iso:
         base: "http://mirror.centos.org/altarch/7/os/ppc64le/"
         updates: "http://mirror.centos.org/altarch/7/updates/ppc64le/"
         epel: "http://download.fedoraproject.org/pub/epel/7/ppc64le/"
+upgrade_versions:
+ push_repo_url: "git@github.com:open-power-host-os/versions.git"
```
ghost: Same as above.

lib/config.py
```diff
@@ -59,9 +59,12 @@
         dict(help='Base Linux distribution version')
 }
 MOCK_ARGS = {
+    ('--mock-binary',):
+        dict(help='Mock binary path', default='/usr/bin/mock'),
```
ghost: There's no need to have this default here, it's already in the yaml file.

lib/config.py
```diff
@@ -59,9 +59,12 @@
...
+    ('--mock-config',):
+        dict(help='Mock config file', default=''),
```
ghost: Why not omit the `default` parameter and leave the default value `None`?

lib/iso_spinner.py
```diff
@@ -114,13 +114,13 @@ def _create_spin_kickstart(self):
...
-                    package_group_list.append(group)
-            for package_group in package_group_list:
-                f.write(package_group + "\n")
+            groups = self.config.get('automated_install_package_groups')
+            for hostos_group in self.config.get('hostos_packages_groups').keys():
+                hostos_group = "@%s" % hostos_group
+                if hostos_group not in groups:
+                    groups.append(hostos_group)
+            for group in groups:
+                f.write(group + "\n")
```
ghost, line 129: Regarding commit message:
"Also, implemented a related small variable renaming in package group creation code
in ISO creator."
I'd write something simpler:
"Also, renamed related variables in ISO creator."

config.yaml
```diff
@@ -80,10 +80,10 @@ build_iso:
         - "@remote-system-management"
         - "@virtualization-platform"
         - "grub2"
-    mock_spin_repo:
+    mock_iso_repo:
         name: "hostos"
         dir: "/host-os-repo"
-    distro_repo_url:
+    distro_repo_urls:
```
ghost: `distro_repos_urls`

lib/config.py
```diff
@@ -100,6 +100,25 @@
     ('--packages-dir', '-d'):
         dict(help='Directory of packages used in the ISO image.',
              default='./result'),
+    ('--iso-name',):
+        dict(help='ISO name.',
+             default='OpenPOWER-Host_OS'),
```
ghost: There's no need to have those defaults here. They are already in the config.yaml file.

lib/config.py
```diff
@@ -100,6 +100,25 @@
...
+    ('--log-file',):
+        dict(help='ISO name.',
+             default='/var/log/host-os/iso.log'),
+    ('--kickstart-file',):
+        dict(help='kickstart file path',
```
ghost: Most help texts start with uppercase.

lib/config.py
```diff
@@ -188,47 +189,55 @@ def _setup_config_parser_options(self, subcommands):
...
 
+
     def parse_config_file(self, config_file_path):
         """
-        Parse the configuration file and return a dictionary containing the
-            parsed values.
+        Parse configuration from a configuration file
+
+        Args:
+            config_file_path(str): configuration file path
```
ghost: You could mention the format of the configuration file (yaml)

lib/config.py
```diff
@@ -188,47 +189,55 @@ def _setup_config_parser_options(self, subcommands):
...
+
+        Returns:
+            dict: Configuration options provided by user in configuration file.
```
ghost: I'd write: "options set by the configuration file"
The user might not have modified the file.

lib/config.py
```diff
@@ -188,47 +189,55 @@ def _setup_config_parser_options(self, subcommands):
...
         """
         conf = {}
         with open(config_file_path) as stream:
             conf = yaml.safe_load(stream)
         return conf
 
+
     def parse(self):
+        """
+        Parse configuration
```
ghost: This doesn't say much. What about:
"Parse configuration file and command line arguments"

lib/build_manager.py
```diff
@@ -47,6 +49,48 @@ def __call__(self):
...
+        Args:
+            distro (Distribution): Linux distribution
+            packages ([str]): packages names
+        """
+
+        # create package builder based on distro
+        if distro.lsb_name == "CentOS":
+            mock_config_file = CONF.get('build_packages').get('mock_config').get(
+                distro.lsb_name).get(distro.version)
+            package_builder = mockbuilder.Mock(mock_config_file)
```
ghost: I don't know if we should keep this distro specifics here, but this looks like a minor drawback from having the build functions removed from the distro class.

config.yaml
```diff
@@ -1,5 +1,6 @@
 common:
  log_file: "/var/log/host-os/builds.log"
+ log_size: 2097152
```
ghost: If you're adding it here, it can be removed from config.py.
You could actually remove all defaults from there since now everything is in the yaml file.

ghost (changes requested):

lib/config.py
alanoe: This is a leftover from a previous implementation idea, fixing it

lib/config.py
alanoe: I prefer to remove all the default values from the command line parser in a separate patch

lib/config.py

lib/config.py

lib/config.py
alanoe: Leftover from a previous implementation idea, fixing it

lib/config.py
alanoe: I prefer to remove all the default values from the command line parser in a separate patch

lib/config.py

lib/config.py

lib/config.py

lib/package.py

config.yaml
alanoe: I prefer to remove all the default values from the command line parser in a separate patch

config.yaml
alanoe: Ok. This directory name makes sense with the current scope of the directory. If the scope is updated with @gutoyr 's patch, he may rename it.

I'm wondering if "package_repo" may be misunderstood with a yum repository...

config.yaml
alanoe: Done

lib/build_manager.py
alanoe: There is a 1-N relationship between package builder and distro. Handling this logic in the distro class or a third class are the possible solutions. As the number of distros we'll support is small and the logic to infer the package builder from the distro is very simple, IMHO keeping this logic in a third class (in this case, build manager) is simpler and clearer.

lib/iso_spinner.py
alanoe: Done

config.yaml
alanoe: I do not understand why we should not push directly to the main repository. In the case of our team, that's exactly what we want to do. You are afraid that a member of our team will run "host_os.py --build-release-notes" accidentally? If that is a possibility, why forcing him to use "host_os.py --build-release-notes --push-repo-url="git@github.com:open-power-host-os/open-power-host-os.github.io.git"  is safer?

In the case of someone outside our team, he/she will not have permission to do so, as you mentioned, so there is no issue IMHO.

ghost: We always push to the main repository through PRs, not directly. We may want to push directly automatically in our weekly builds in the future, but that can be explicitly set in the weekly job.

Someone may be testing the commands and not realize the default is to push to the main repository. Forcing to use `--push-repo-url` makes the user think (or, at least, look) about where he wants to push.

config.yaml
```diff
@@ -22,6 +20,9 @@ default:
  commit_updates: True
  push_updates: True
  force_rebuild: False
+release_notes:
+ release_notes_repo_url: "https://github.com/open-power-host-os/open-power-host-os.github.io.git"
+ release_notes_repo_branch: "master"
```
farosas: These names can now be simply `repo_url` and `repo_branch`

config.yaml
farosas: The real problem here is the `build` in the name of the key, which is very confusing. Other than that I think it's a matter of defining if we want to be as generic as possible and lose usability or if we don't care having some 'domain language' for the user to learn. In this particular case, I think the latter is simpler and it will save us time.

lib/config.py
```diff
@@ -29,7 +29,7 @@
         dict(help='Packages metadata git repository URL'),
     ('--packages-metadata-repo-branch',):
         dict(help='Packages metadata git repository branch'),
-    ('--build-versions-repo-dir',):
+    ('--packages-metadata-repo-target-path',):
```
farosas: This patch could be squashed with the previous one

lib/config.py
```diff
@@ -251,7 +251,7 @@ def parse(self):
         # Each subcommand may have a node for specific configurations
         # at the same level of the 'default' node
         COMMAND_TO_CONFIG_NODE = {
-            "build-iso": "iso",
+            "build-iso": "build_iso",
```
farosas: We could now simplify this implementation using a list of sub-command names.

tests/unit/config_tests.py
```diff
@@ -17,7 +17,7 @@ class TestConfigParser(unittest.TestCase):
         (['build-packages', '--packages=foo'], 'packages', ['foo']),
         (['build-packages', '--result-dir=foo'], 'result_dir', 'foo'),
         (['build-packages', '--packages-repos-target-path=foo'], 'packages_repos_target_path', 'foo'),
-        (['build-packages', '--keep-builddir'], 'keep_builddir', True),
+        (['build-packages', '--keep-build-dir'], 'keep_builddir', True),
```
farosas: The second element of the tuple should change too.

tools/build_release_notes.py
```diff
@@ -144,8 +144,7 @@ def push_website_head_commit(
 def run(CONF):
     versions_repo = setup_versions_repository(CONF)
 
-    packages_names = (CONF.get('common').get('packages')
-                      or packages_manager.discover_packages())
+    packages_names = packages_manager.discover_packages()
```
farosas: What do you mean with 'are never used'? And what about command line usage?

config.yaml
```diff
@@ -35,7 +35,9 @@ build_release_notes:
 build_iso:
     iso_name: "OpenPOWER-Host_OS"
     log_file: "/var/log/host-os/iso.log"
-    mock_config: "./mock_configs/CentOS/7/spin-iso-CentOS-7-ppc64le.cfg"
+    mock_config:
+     CentOS:
+      '7': "./mock_configs/CentOS/7/spin-iso-CentOS-7-ppc64le.cfg"
```
farosas: side note: why don't we infer the path from the yaml keys?

lib/config.py
```diff
@@ -100,6 +100,25 @@
     ('--packages-dir', '-d'):
         dict(help='Directory of packages used in the ISO image.',
              default='./result'),
+    ('--iso-name',):
+        dict(help='ISO name.',
+             default='OpenPOWER-Host_OS'),
+    ('--log-file',):
+        dict(help='ISO name.',
+             default='/var/log/host-os/iso.log'),
+    ('--kickstart-file',):
```
farosas: This name changed.

lib/config.py
```diff
@@ -100,6 +100,25 @@
...
+        dict(help='Path of a kickstart file, used to automate the installation of a RPM-based Linux distribution',
+             default='host-os.ks'),
+    ('--hostos-packages-groups',):
+        dict(help='Packages groups in yum repository'),
+    ('--automated-install-package-groups',):
+        dict(help='Packages and package groups which are installed using automated installation'),
+    ('--mock_iso_repo.name',):
```
farosas: Underscores and dot instead of dashes

lib/config.py
```diff
@@ -100,6 +100,25 @@
...
+        dict(help='Name of the yum repository which will be created with OpenPOWER Host OS packages'),
+    ('--mock_iso_repo.dir',):
```
farosas: Same here

lib/package_builder.py
```diff
@@ -34,3 +30,6 @@ def clean(self):
         """
         clean up steps.
         """
+
+
+
```
farosas: whitespace

config.yaml
alanoe: Thanks, this is clearer to me now. I thought about using the URL of our bot user git repo as default here. However, as I discussed with Olav today, running update-versions and build-release-versions subcommands won't work out of the box anyway because the user must have the private SSH key of the GitHub user and a SSH config setup in the system running host_os.py. IMHO either command line should parametrize this or at least we should document this so that these commands work in the command line and we do not have a dependency with Jenkins here, but Olav told me this has low priority. I'll just leave the attribute blank for now.

lib/config.py
```diff
@@ -43,6 +43,9 @@
     ('--packages-repos-target-path', '-R'):
         dict(help='Directory where to clone code repositories',
              default='/var/lib/host-os/repositories'),
+    ('--no-update-packages-repos-target-path',):
```
farosas: name does not match the config key/description

farosas (dismissed):
Aside from what's inline, a couple of points: 

There are too many commits in this PR and some of them are not related to the proposed change, this makes it hard to review and know which changes we already agreed on and which are new. I suggest that in the future you try to map PRs to trello tasks to make life easier for everyone.

I would also dedicate a separate PR for patches which are simply renaming things, because they tend to bring a lot of discussion and we don't want them to delay the "feature" patches or compromise the overall review process.

config.yaml
alanoe: release-notes subcommand may accept a packages_metadata_repo_url (versions.git) arg and a release_notes_repo_url (open-power-host-os.github.io.git) arg. I thought that calling one of them repo_url would be a little confusing for the command line user, but up to you guys.

tools/build_release_notes.py
alanoe: We only use this subcommand to create release notes of all packages. What is the usefulness of doing it for a subset?

tests/unit/config_tests.py
alanoe: Done

lib/config.py

lib/config.py

config.yaml
alanoe: +1. We can improve it in a separate PR.

lib/config.py

lib/config.py

tools/build_release_notes.py
farosas: I don't know, just thought we would like to keep consistent with build-packages. Ok then.

lib/config.py
alanoe: I'll do it in a separate patch

lib/config.py

lib/package_builder.py
alanoe: Done

config.yaml
ghost: I think "packages metadata" is too generic. The new user will not have any clue what is this metadata. By having "versions" in the name, he at least knows what repo it refers to. It also makes explicit *only* the versions repo (or a fork) can be used here.
Having to write "packages metadata" in every variable and documentation, in my opinion, just adds extra words and complexity with no gain of value.

alanoe: OH> I think "packages metadata" is too generic.

But that's exactly what this git repository holds, packages metadata! Any other term I can think of is imprecise and unclear. You think "versions" is a better name? "versions" is a confusing name because it seems like this git repo keeps only versions number and it keeps much more than that. "packages" is also a little confusing because people may confuse with the packages source code git repositories (it still is better than 'versions', though). I prefer a longer clearer name.

btw nothing prevents us from renaming the 'versions' git repository to 'packages' or 'packages_metadata'

side-note: "packages metadata" is a term already used in parts of the code, this is one way to make the terminology more consistent.

lib/config.py
```diff
@@ -158,16 +158,23 @@ def __init__(self):
         # create the top-level parser
         self.parser = argparse.ArgumentParser()
         self._CONF = None
-        self._setup_config_parser_options()
+        self._setup_config_parser_options(SUBCOMMANDS)
```
ghost: I'd rewrite the commit message to make the main point clearer. It is not the method rename, it is the addition of the parameter. Also, please try to keep the short commit messages short. It is really annoying to read messages like this:
```
Rename subparsers manager from subparser to subparsers to make code e…

…asier to understand
```
It also only appears partially in the commits page. I suggest:
```
Add "subcommands" parameter to ConfigParser methods

This restricts knowledge of external constants to a single method in the class.
Also, rename the "_add_subparser" method and document related methods.
```

lib/config.py
```diff
@@ -28,8 +28,7 @@
         dict(help='Packages metadata git repository branch'),
     ('--packages-metadata-repo-target-path',):
         dict(help='Directory to clone the packages metadata git repository. '
-             'A subdirectory with the name of the git repository will be created here',
-             default='.'),
+             'A subdirectory with the name of the git repository will be created here'),
```
ghost: Commit message: mentions those defaults are already in config.yaml

ghost (changes requested):
I agree completely with Fabiano's previous comment. This PR is already too big as it is. It has a lot of unrelated refactorings and new commits keep appearing.
Unfortunately, we are stuck to GitHub's review interface, so every time you rebase, I don't know if you added new commits, modified the old ones, what changed...
I suggest keeping this PR with commits until "Add log_size and verbose options to config.yaml" and opening new PRs to the newer ones.

Scheduler and BuildManager improvements have nothing to do with the config parser and belong in a separate PR:
- Replace __call__() with a method named with a verb in classes 
- Improve Scheduler class documentation

Passing package options in the command line is a different feature and deserves proper review of its own:
-  Support overriding SCM repository URL, commit ID and branch in comman… 

I believe the remaining commits belong to PR #196

alanoe:
OH> I suggest keeping this PR with commits until "Add log_size and verbose options to config.yaml" and opening new PRs to the newer ones.

Opened:
https://github.com/open-power-host-os/builds/pull/197
https://github.com/open-power-host-os/builds/pull/198
https://github.com/open-power-host-os/builds/pull/199
https://github.com/open-power-host-os/builds/pull/200

I'll open more PRs when those are merged to avoid wasting time with infinite rebases. Could you guys pls review those ASAP?

lib/config.py

Issue #186, opened 2017-02-14
README suggests possibly sub-optimal steps (e.g., SMT off for builds too, instead of only to run KVM guests)


mfoliveira:
Currently, the README.rst file (landing page for this project [1]) suggests to set SMT off in the general case (e.g., build and run).

> Settings
> Disable SMT
> $ sudo ppc64_cpu --smt=off

Indeed, SMT off is required to run KVM guests. 
However, that's not a technical limitation when building the packages, and actually SMT on would help with the build speed if multiple make jobs are being used (i.e., `make -jN`).

So, it would be great to take advantage of more threads (if the package build scripts/SPECs don't already do it), and clarify the scenarios that (don't ) need SMT off.

[1] https://github.com/open-power-host-os/builds/

fabianorosas:
Yes, the documentation certainly needs to be made more clear in that aspect. Thank you for taking the time to report this.

For completeness, here's a complementary discussion: https://github.com/open-power-host-os/builds/issues/16

mfoliveira:
@fabianorosas okay, thank you.

mfoliveira:
@fabianorosas noticed something else in README that I believe we could improve in the benefit of our users / user experience; so I just changed this issue's title to be more general (rather than open a new one for a minor thing. Let me know if you'd prefer it otherwise).

Any reason why the installation of `epel-release` is split into 2 commands?

```
$ sudo yum install --downloadonly --downloaddir=. epel-release
$ sudo yum localinstall epel-release-7-5.noarch.rpm # Note the version may change.
```

It's possible to install it with:
```
$ sudo yum install epel-release
```

It has the advantage of being version independent, thus not requiring to have users to read/care about a caveat note on the version number.
Which, by the way, happened to change version numbers to 7-6, so the single-command variant would help with that  :)

```
# rpm  -q epel-release
epel-release-7-6.noarch
```

mfoliveira:
This would be much better in usability/ux terms in the form a meta-package; e.g., `host-os-build-deps`

```
$ sudo yum install -y $(cat rpm_requirements.txt)
```

mfoliveira:
And this too, but totally agree with the point here:

> We decided that's is a little bit intrusive to install all the produced RPMs since some of them may not fit your needs. [snip]

However, again in the benefit of our users, I'd suggest to make _multiple_ meta-packages at different levels of dependencies installed.
[...]

Issue #168, opened 2017-01-31
Issue when trying to run build-package


tcveritas:
I'm running into an issue when trying to run the build-package command. This is after I created a user "admin" and su - to it 

![build-package](https://cloud.githubusercontent.com/assets/25467720/22485593/db4c2a4e-e7cb-11e6-9a64-a3a9de2b3ce7.PNG)

tcveritas:
I was able to move past this issue thanks to the response in issue #166. However, I am now running into the following error. This is my first go-around using github, so any help would be appreciated 
![nexterror](https://cloud.githubusercontent.com/assets/25467720/22485931/14f9a220-e7cd-11e6-90c0-84a787807e05.PNG)

gutoyr:
Seems that the user used to clone open-power-host-os/builds repository is not the same as the one that executed `python host_os.py build-package`.

When building the packages, host_os.py tries to write inside open-power-host-os/builds repository, therefore, the user that executes `python host_os.py build-package` must be the same as the one passed to `sudo python host_os.py set-env --user <my_user>`.

The easiest way to fix it is starting from the beginning, following this steps:

1. log as **user1**
2. as **user1**, clone open-power-host-os/builds repository
`git clone https://github.com/open-power-host-os/builds.git`
3. set build enviroment for **user1**
`sudo python set-env --user user1`
4. as **user1**, build the packages
`python build-package`

This information is missing in the project documentation, I will address it.

tcveritas:
That helped me make progress, but it still failed out. Looks like it's failing on trying to find a RHEL repo? 
![issue3](https://cloud.githubusercontent.com/assets/25467720/22514625/8b93e79a-e865-11e6-9247-13d090874cdd.PNG)

gutoyr:
svn failed to checkout gcc repository, probably due to some sort of network issue. You can try to do it manually before build the packages:

`svn checkout svn://gcc.gnu.org/svn/gcc/branches/redhat/gcc-4_8-branch /var/lib/host-os/repositories/gcc`

then:

`python host_os.py build-package`

If the _svn checkout_ fails, then it is a network issue. In this case, you can try what is suggested by gnu/gcc [1] under "Using the SVN repository" topic and  replace "svn://" with "http://".

[1] https://www.gnu.org/software/gcc/svn.html

tcveritas:
Still seems to be throwing the same error after following the process above 
[...]

PR #191, opened 2017-02-20
Changes on set-env, work-dir and result-dir

gyribeiro:
- Remove set-env subcommand
- Add work-dir support
- Change result-dir behavior


host_os.py
```diff
@@ -54,14 +53,15 @@
               "script: %s" % ", ".join(missing_packages))
         sys.exit(MISSING_PACKAGES_ERROR)
 
-    if os.getuid() is 0 and subcommand != 'set-env':
-        print("Please, do not run this command as root, run "
-              "host_os.py set-env --user <YOUR_USER_LOGIN> command in order to "
-              "properly setup user and directory for build scripts")
+    if os.getuid() is 0:
+        print("Please, do not run this command as root.")
```
ghost, line 64: Unrelated to this patch: I believe we should use `LOG.error` here instead of print.

host_os.py
```diff
@@ -54,14 +53,15 @@
...
         sys.exit(TOO_MUCH_PRIVILEGE_ERROR)
 
-    if os.getuid() is not 0 and subcommand == 'set-env':
-        print("The set-env command should be run with root privileges")
+    mock_users = grp.getgrnam('mock').gr_mem
+    user = os.environ['USER']
+    if user not in mock_users:
+        print("User must be in mock group, please run "
+              "sudo usermod -a -G mock $(whoami)")
```
ghost: Enclose the command in single quotes to separate it from text.

config.yaml
```diff
@@ -1,5 +1,6 @@
 default:
  branch: 'powerkvm-v3.1.1'
+ work_dir: 'build'
```
ghost: About commit message:
artifects -> artifacts
I'd mention those are temporary/auxiliary artifacts and not necessarily final results.

lib/iso_spinner.py
```diff
@@ -133,6 +134,9 @@ def _spin(self):
         self._run_mock_command("--shell '%s'" % spin_cmd)
 
     def _save(self):
+        utils.create_directory(self.result_dir)
+        os.chmod(self.result_dir, 0777)
```
ghost: Why does it need full access?
I've sent a patch that changes this in the build-packages command: https://github.com/open-power-host-os/builds/pull/177/commits/a33488c55bf93c2c4d1739a433f991f83984c282

config.yaml
```diff
@@ -1,6 +1,7 @@
 default:
  branch: 'powerkvm-v3.1.1'
  work_dir: 'build'
+ result_dir: 'result'
```
ghost: I prefer actions instead of state description in the commit short message:
"Make result-dir a global CLI parameter"
Using future also looks like it is not yet valid for this commit:
"It is now applicable to all sub-commands, not just build-package."

lib/config.py
```diff
@@ -280,7 +277,9 @@ def setup_default_config():
         print("Failed to parse settings")
         sys.exit(2)
 
-    log_helper.LogHelper(log_file_path=CONF.get('default').get('log_file'),
+    log_file = 'builds.log'
```
ghost: I'd name this `LOG_FILE_NAME`

lib/log_helper.py
```diff
@@ -34,15 +34,14 @@ def __init__(self, log_file_path=None, verbose=False, rotate_size=0):
         sh.setFormatter(logging.Formatter('%(message)s'))
         logger.addHandler(sh)
 
-        if log_file_path:
```
ghost: If you're removing this `if`, then make the `log_file_path` required. Having a default value that breaks the function makes no sense.

lib/package.py
```diff
@@ -74,9 +74,12 @@ def __init__(self, name, force_rebuild=True):
         self.repository = None
         self.build_files = None
         self.download_build_files = []
-        locks_dir = CONF.get('default').get('repositories_path')
+        repositories_dir = 'repositories'
```
ghost: I'd name this constant `REPOSITORIES_DIR_NAME`

lib/config.py
```diff
@@ -192,6 +192,10 @@ def _setup_config_parser_options(self):
         self.parser.add_argument('--log-size',
                                  help='Size in bytes above which the log file '
                                  'should rotate', type=int, default=2<<20)
+        self.parser.add_argument('--work-dir', '-w',
+                                 help='Directory used to store all artifacts '
+                                 'created during the process.',
+                                 default='build')
```
ghost: The default is already in `config.yaml`, there's no reason to have it here too. This will soon be fixed for other parameters too.

lib/mockbuilder.py
```diff
@@ -148,7 +148,8 @@ def _install_external_dependencies(self, package):
 
     def _create_build_directory(self, package):
         self.build_dir = os.path.join(
-            os.getcwd(), 'build', self.timestamp, package.name)
+            os.path.abspath(CONF.get('default').get('work_dir')), 'build',
+            self.timestamp, package.name)
```
ghost: About the commit message: I suggest a more direct language:
Place the build directory used by mock under work-dir

lib/package.py
```diff
@@ -32,6 +33,9 @@
 
 BUILD_CACHE_DIR = os.path.join(CONF.get('default').get('work_dir'), "cache")
 
+RESULT_DIR = os.path.join(CONF.get('default').get('result_dir'),
+                          datetime.datetime.now().isoformat())
```
ghost: This timestamp will potentially be different from the one created in `mockbuilder.py` and used as the mock's build directory name. It will then be hard to match a build directory with a corresponding results directory when debugging.
I think we could remove the `build_results_dir` attribute from here and create that path from the package name and `Mock.timestamp` in `Mock.copy_results()`, which is the only place this attribute is used.

ghost (changes requested):

lib/mockbuilder.py
gyribeiro: Done

lib/package.py
gyribeiro: Done

lib/package.py
gyribeiro: Done

lib/log_helper.py
gyribeiro: Done

lib/config.py
gyribeiro: Done

lib/config.py
gyribeiro: Due to how unit tests were implemented, this is the only way to test default values. I suggest to open a task to refactor the unit test implementation to start considering default values provided by the config.yaml, as well.

lib/iso_spinner.py
gyribeiro: Removed

host_os.py
gyribeiro: Done

host_os.py
gyribeiro: For this specific scenario, user must receive a visual feedback when fails to run host_os.py as root. Forcing the user to check the logs just to figure out what is wrong, is counterproductive for this case .

config.yaml
gyribeiro: Done

config.yaml
gyribeiro: Done

host_os.py
```diff
@@ -56,14 +55,15 @@
...
+        print("Please, do not run this command as root.")
         sys.exit(TOO_MUCH_PRIVILEGE_ERROR)
 
-    if os.getuid() is not 0 and subcommand == 'set-env':
-        print("The set-env command should be run with root privileges")
+    mock_users = grp.getgrnam('mock').gr_mem
+    user = os.environ['USER']
+    if user not in mock_users:
+        print("User must be in mock group, please run "
+              "'sudo usermod -a -G mock $(whoami)'")
```
alanoe: Having the user in the 'mock' group is required for running subcommands which do not use mock (eg release-notes, upgrade-versions and update-versions-readme) ? If not, why block the user from executing them?

README.rst
```diff
@@ -48,21 +48,11 @@ Settings
 $ git clone https://github.com/open-power-host-os/builds.git
```
alanoe, line 48: Pls add to commit longer msg why it was decided to delete this subcommand.

lib/config.py
```diff
@@ -200,6 +200,10 @@ def _setup_config_parser_options(self):
         self.parser.add_argument('--log-size',
                                  help='Size in bytes above which the log file '
                                  'should rotate', type=int, default=2<<20)
+        self.parser.add_argument('--work-dir', '-w',
+                                 help='Directory used to store all artifacts '
```
alanoe: - I'd avoid using Jenkins terminology (artifacts) in host_os.py
- Make it explicit this directory contains *temporary* files

I suggest "artifacts" => "temporary files"

lib/config.py
```diff
@@ -24,6 +24,7 @@
 from lib import utils
 
 LOG = logging.getLogger(__name__)
+LOG_FILE_NAME = 'builds.log'
```
alanoe: IMHO it makes more sense to have this in log_helper.py than here

lib/package.py
```diff
@@ -32,6 +32,7 @@
 
 # TODO: make this configurable by a base dir parameter
 BUILD_CACHE_DIR = "cache"
+REPOSITORIES_DIR_NAME = 'repositories'
```
alanoe: What repositories? I prefer PACKAGES_REPOS_TARGET_PATH , this is the name I used in PR builds#190

config.yaml
alanoe: IMHO tmp/temp/workspace are more intuitive names than "build", but up to you guys

lib/config.py
```diff
@@ -145,7 +141,7 @@ def discover_packages():
     versions_repo_url = config.get('build_versions_repository_url')
     versions_repo_name = os.path.basename(os.path.splitext(versions_repo_url)[0])
     build_versions_repo_dir = os.path.join(
-        config.get('build_versions_repo_dir'),
+        config.get('work_dir'),
```
alanoe, line 147: IMHO it is better to keep this in the "repositories" directory along with the packages repositories

lib/package.py
```diff
@@ -30,8 +30,7 @@
 CONF = config.get_config().CONF
 LOG = logging.getLogger(__name__)
 
-# TODO: make this configurable by a base dir parameter
-BUILD_CACHE_DIR = "cache"
+BUILD_CACHE_DIR = os.path.join(CONF.get('default').get('work_dir'), "cache")
 REPOSITORIES_DIR_NAME = 'repositories'
```
alanoe: Why keep the cache dir full path and repositories dir name? Be consistent, either use full paths in both or dir names in both.

lib/mockbuilder.py
```diff
@@ -149,7 +149,8 @@ def _install_external_dependencies(self, package):
 
     def _create_build_directory(self, package):
         self.build_dir = os.path.join(
-            os.getcwd(), 'build', self.timestamp, package.name)
+            os.path.abspath(CONF.get('default').get('work_dir')), 'build',
```
alanoe: Currently this will be build/build, this does not make sense. I suggest we use mock_build

config.yaml
alanoe: Wouldn't it be better to allow different directories for package building (eg result/packages) and ISO building (eg result/iso) and not support this argument for other subcommands?

alanoe (changes requested):

lib/config.py
alanoe: I have already done that in PR builds#190

host_os.py
gyribeiro: done

README.rst
gyribeiro: done

lib/config.py
gyribeiro: LogHelper requires the log file name (full path). If this change is required, it should be done in another path.

lib/config.py
gyribeiro: done

lib/config.py
gyribeiro: Make sense but, with the current code structure, this change will introduce a circular dependency which will require a code refactor. Out of the scope of this patch set, if required should be done in another patch.

lib/package.py
gyribeiro: done

lib/package.py
gyribeiro: done

config.yaml
gyribeiro: done

config.yaml
gyribeiro: done

lib/mockbuilder.py
gyribeiro: done

lib/config.py
alanoe: OK. then I suggest we use "packages_repositories" for packages git/svn repositories and "packages_metadata_repository"/"versions_repository" for versions.git (or something similar)

README.rst
```diff
@@ -48,21 +48,11 @@ Settings
...
-Naturally, you need to replace ``LOGIN`` by the user name you'll use
-to run ``host_os.py``, which should not run using root user,
-even if that user doesn't exist yet.
-
-Please see ``--help`` for more options.
-
-::
-
-$ ./host_os.py set-env --help
+$ sudo useradd -a -G mock $(whoami)
```
farosas: `usermod` right?

lib/config.py
```diff
@@ -200,6 +200,10 @@ def _setup_config_parser_options(self):
         self.parser.add_argument('--log-size',
                                  help='Size in bytes above which the log file '
                                  'should rotate', type=int, default=2<<20)
+        self.parser.add_argument('--work-dir', '-w',
+                                 help='Directory used to store all temporary '
+                                 'files created during the process.',
+                                 default='build')
```
farosas: The defaults for work-dir differ.

host_os.py
ghost: `LogHelper` sets the stdout StreamHandler to always display messages with level `info` and above.

lib/package.py
```diff
@@ -30,8 +30,7 @@
 CONF = config.get_config().CONF
 LOG = logging.getLogger(__name__)
 
-# TODO: make this configurable by a base dir parameter
-BUILD_CACHE_DIR = "cache"
+BUILD_CACHE_DIR = os.path.join(CONF.get('default').get('work_dir'), "cache")
```
ghost, line 34: Either mention the build cache directory in the commit message or separate this commit.

lib/mockbuilder.py
```diff
@@ -187,5 +186,6 @@ def copy_results(self, package):
             package(Package): package whose result files will be copied
         """
         package_build_results_dir = os.path.join(
-            self.build_results_dir, package.name)
+            CONF.get('default').get('result_dir'), 'packages',
+            self.timestamp, package.name)
```
ghost, line 191: This will require modifications in the infrastructure repository.
The archiving of artifacts in the `build_host_os` job and the copying of those archives in the `upload_build_artifacts`, `build_host_os_iso` and `upload_iso` jobs will need path changes. Maybe creating a `latest` symlink at the end of the build would make that easier?

ghost (changes requested):

farosas (changes requested):

farosas:
This addresses some of the issues mentioned in #186 , specifically: https://github.com/open-power-host-os/builds/issues/186#issuecomment-280100001, https://github.com/open-power-host-os/builds/issues/186#issuecomment-280101128

farosas:
#168 should be solved by this PR too

lib/config.py
gyribeiro: Done, `versions` is now placed under <work-dir>/<REPOSITORIES_DIR> (REPOSITORIES_DIR is defined on lib/constants.py to avoid circular dependency)

lib/package.py
gyribeiro: done

lib/mockbuilder.py
gyribeiro: done

README.rst
gyribeiro: done

lib/config.py
gyribeiro: done

lib/utils.py
```diff
@@ -191,3 +192,20 @@ def replace_str_in_file(file_path, search, replacement):
         for line in lines:
             line = line.replace(search, replacement)
             f.write(line)
+
+
+def force_symlink(source, link_name):
```
ghost: I'd name the parameters `target_path` and `link_path`

lib/utils.py
```diff
@@ -191,3 +192,20 @@ def replace_str_in_file(file_path, search, replacement):
...
+    """
+    Create a symbolic link pointing to source named link_name, if link_name
+    already exists, then delete it before proceed.
```
ghost: To simplify: "Create a symbolic link to the target path, deleting it beforehand if it already exists."

config.yaml
```diff
@@ -25,7 +25,7 @@ default:
 iso:
     iso_name: "OpenPOWER-Host_OS"
     mock_config: "./mock_configs/CentOS/7/spin-iso-CentOS-7-ppc64le.cfg"
-    packages_dir: "./result"
+    packages_dir: "result/packages/latest"
```
ghost: Commit message: typo in "latest"

ghost (approved):

farosas (approved):

lib/utils.py
gyribeiro: done

lib/utils.py
gyribeiro: done

config.yaml
gyribeiro: done

ghost (approved):

PR #192, opened 2017-02-20
Remove build_results_dir attribute from Package class

ghost:
This attribute is inferred from an option that is only available in the
build-package command and so it breaks the upgrade-versions command.

Since the only user of this attribute is the Mock class, that copies
the actual build results to this directory, it makes sense to create
this path there.


gyribeiro (approved):

olavphibm:
start build

farosas (approved):

PR #194, opened 2017-02-22
Update RPM dependencies

alanoe:


alanoe:
start build

farosas (approved):

rpm_requirements.txt
```diff
@@ -7,4 +6,3 @@ nosync
 python-lxml
 PyYAML
 subversion
-wget
```
ghost: Don't we need those because of backwards compatibility? We should at least mention them in the building old versions documentation.

ghost (approved):

rpm_requirements.txt
alanoe: Good point, I'll add that to the documentation

PR #195, opened 2017-02-22
Fix copying of RPMs during ISO creation

ghost:
The source RPMs were not being copied to the chroot because
subdirectories were added to the results directory in commit
f20d7166870e918c3976fb6846feca688ea4df44.


alanoe (approved):

alanoe (changes requested):
Remove unused glob import

gyribeiro (approved):

PR #196, opened 2017-02-22
Create config single definition

alanoe:
Depends on builds/#200 and builds/#207

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -18,9 +18,6 @@ build_packages:
  packages:
  keep_build_dir: False
  force_rebuild: False
- mock_config:
-  CentOS:
-   '7': "./mock_configs/CentOS/7/CentOS-7-ppc64le.cfg"
  mock_args:
    # If the system has less RAM than required, the plugin will be disabled
    "--enable-plugin=tmpfs
@@ -36,9 +33,6 @@ build_release_notes:
  push_repo_branch: "master"
 build_iso:
     iso_name: "OpenPOWER-Host_OS"
-    mock_config:
-     CentOS:
-      '7': "./mock_configs/CentOS/7/build-iso-CentOS-7-ppc64le.cfg"
     mock_args:
       # If the system has less RAM than required, the plugin will be disabled
       "--enable-plugin=tmpfs
--- a/config.yaml
+++ b/config.yaml
@@ -1,87 +1,87 @@
 common:
- work_dir: 'workspace'
- result_dir: 'result'
- verbose: False
- log_size: 2097152
- packages_metadata_repo_url: "https://github.com/open-power-host-os/versions.git"
- packages_metadata_repo_branch: 'master'
- http_proxy: ''
- distro_name: 'CentOS'
- distro_version: '7'
- architecture: 'ppc64le'
- mock_binary: "/usr/bin/mock"
- commit_updates: True
- push_updates: True
- updater_name:
- updater_email:
+  work_dir: 'workspace'
+  result_dir: 'result'
+  verbose: False
+  log_size: 2097152
+  packages_metadata_repo_url: "https://github.com/open-power-host-os/versions.git"
+  packages_metadata_repo_branch: 'master'
+  http_proxy: ''
+  distro_name: 'CentOS'
+  distro_version: '7'
+  architecture: 'ppc64le'
+  mock_binary: "/usr/bin/mock"
+  commit_updates: True
+  push_updates: True
+  updater_name:
+  updater_email:
 build_packages:
- packages:
- keep_build_dir: False
- force_rebuild: False
- mock_args:
-   # If the system has less RAM than required, the plugin will be disabled
-   "--enable-plugin=tmpfs
-   --plugin-option=tmpfs:keep_mounted=True
-   --plugin-option=tmpfs:max_fs_size=32g
-   --plugin-option=tmpfs:required_ram_mb=39800
-   --verbose"
- update_packages_repos_before_build: True
+  packages:
+  keep_build_dir: False
+  force_rebuild: False
+  mock_args:
+    # If the system has less RAM than required, the plugin will be disabled
+    "--enable-plugin=tmpfs
+    --plugin-option=tmpfs:keep_mounted=True
+    --plugin-option=tmpfs:max_fs_size=32g
+    --plugin-option=tmpfs:required_ram_mb=39800
+    --verbose"
+  update_packages_repos_before_build: True
 build_release_notes:
- release_notes_repo_url: "https://github.com/open-power-host-os/open-power-host-os.github.io.git"
- release_notes_repo_branch: "master"
- push_repo_url:
- push_repo_branch: "master"
+  release_notes_repo_url: "https://github.com/open-power-host-os/open-power-host-os.github.io.git"
+  release_notes_repo_branch: "master"
+  push_repo_url:
+  push_repo_branch: "master"
 build_iso:
-    iso_name: "OpenPOWER-Host_OS"
-    mock_args:
-      # If the system has less RAM than required, the plugin will be disabled
-      "--enable-plugin=tmpfs
-      --plugin-option=tmpfs:keep_mounted=True
-      --plugin-option=tmpfs:max_fs_size=32g
-      --plugin-option=tmpfs:required_ram_mb=39800
-      --verbose"
-    packages_dir: "result/packages/latest"
-    automated_install_file: "host-os.ks"
-    installable_environments:
-        OpenPOWER Host OS all:
-            - "open-power-host-os-all"
-        OpenPOWER Host OS base:
-            - "open-power-host-os-base"
-        OpenPOWER Host OS container:
-            - "open-power-host-os-container"
-        OpenPOWER Host OS virtualization:
-            - "open-power-host-os-virt"
-        OpenPOWER Host OS virtualization management:
-            - "open-power-host-os-virt-management"
-        OpenPOWER Host OS RAS:
-            - "open-power-host-os-ras"
-    base_distro_minimal_install_groups:
-        - "core"
-    iso_root_fs_packages_groups:
-        - "base"
-        - "core"
-        - "debugging"
-        - "virtualization-hypervisor"
-        - "virtualization-tools"
-        - "network-file-system-client"
-        - "remote-system-management"
-        - "virtualization-platform"
-    iso_root_fs_packages:
-        - "grub2"
-    mock_iso_repo_name: "host_os"
-    mock_iso_repo_dir: "/host-os-repo"
-    distro_repos_urls:
-        base: "http://mirror.centos.org/altarch/7/os/ppc64le/"
-        updates: "http://mirror.centos.org/altarch/7/updates/ppc64le/"
-        extras: "http://mirror.centos.org/altarch/7/extras/ppc64le/"
-        epel: "http://download.fedoraproject.org/pub/epel/7/ppc64le/"
+  iso_name: "OpenPOWER-Host_OS"
+  mock_args:
+    # If the system has less RAM than required, the plugin will be disabled
+    "--enable-plugin=tmpfs
+    --plugin-option=tmpfs:keep_mounted=True
+    --plugin-option=tmpfs:max_fs_size=32g
+    --plugin-option=tmpfs:required_ram_mb=39800
+    --verbose"
+  packages_dir: "result/packages/latest"
+  automated_install_file: "host-os.ks"
+  installable_environments:
+    OpenPOWER Host OS all:
+    - "open-power-host-os-all"
+    OpenPOWER Host OS base:
+    - "open-power-host-os-base"
+    OpenPOWER Host OS container:
+    - "open-power-host-os-container"
+    OpenPOWER Host OS virtualization:
+    - "open-power-host-os-virt"
+    OpenPOWER Host OS virtualization management:
+    - "open-power-host-os-virt-management"
+    OpenPOWER Host OS RAS:
+    - "open-power-host-os-ras"
+  base_distro_minimal_install_groups:
+  - "core"
+  iso_root_fs_packages_groups:
+  - "base"
+  - "core"
+  - "debugging"
+  - "virtualization-hypervisor"
+  - "virtualization-tools"
+  - "network-file-system-client"
+  - "remote-system-management"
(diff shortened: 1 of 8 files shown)


config_metadata.yaml
```diff
@@ -0,0 +1,169 @@
+common:
```
farosas, line 1: Should this file be hidden to avoid inadvertent edits?

lib/config.py
```diff
@@ -155,47 +46,61 @@ def __init__(self):
...
         subparsers = self.parser.add_subparsers(
             dest="subcommand",
             help="Available subcommands")
-        for command, help_msg, arg_groups in subcommands:
-            command_parser = subparsers.add_parser(command, help=help_msg)
-            for arg_group in arg_groups:
-                for arg, options in arg_group.items():
-                    command_parser.add_argument(*arg, **options)
+
+        # setup parser from config metadata file
```
farosas: Why do we need to infer type, nargs, action? Wasn't that the supposition when we *didn't* have a config_metadata file? Why not simply keep these explicitly as well as any other `add_argument` argument? I think it would simplify the code below, avoid any bugs in the inferring logic and also allow us to use other arguments.

Or am I forgetting something here?

config_metadata.yaml
alanoe: I don't know if it is necessary, given that the file content YAML structure is different from config.yaml. Alternatively, there could be a warning at the top of the file. Up to you guys.

lib/config.py
alanoe: I am not sure I understood you. Default action is 'store', somewhere we must explicitly define the action if it is a different value (eg store_true or store_false). Same for nargs. The difference is that before we had those things defined for *each argument* in the SUBCOMMANDS constants, now they are inferred for *all arguments* from the config metadata file. The idea here is to automatically create the command line parser from the config metadata file, there should be no code specific to one argument.

lisibann (commented):
We need some documentation related to the config metadata file and the execution of sync_config_yaml.py.

lib/config.py
farosas: My suggestion is to explicitly set these values in config_metadata now that this won't be edited by the user:
```
verbose: 
  argparse_args:
    help:
    default:
    action:
    const:
    choices:
    required:
    metavar: 
    dest:
```

Just bringing to your attention that the current design was based on premises that are not true anymore, that's all.

farosas (approved):
Reviewed only b5bf78a28f7999e044deede23d247f52108a7e1a as requested

lib/config.py
alanoe: I vote for the design proposed in this patch, which infers most of these argparse fields and requires us to keep a simpler and smaller config metadata file. I still do not understand the benefit of mentioning those explicitly.

config.yaml
```diff
@@ -1,88 +1,85 @@
...
- arch_and_endianness: 'ppc64le'
- mock_binary: "/usr/bin/mock"
- commit_updates: True
- push_updates: True
- updater_name:
- updater_email:
+build_iso:
+  automated_install_file: host-os.ks
+  automated_install_packages_groups:
+  - '@base'
```
ghost: Shouldn't this be at a higher indentation level?

lib/config.py
```diff
@@ -149,46 +49,61 @@ def __init__(self):
...
-                                 help='Set the scripts to be verbose',
-                                 action='store_true')
-        self.parser.add_argument('--log-size',
-                                 help='Size in bytes above which the log file '
-                                 'should rotate', type=int)
-        self.parser.add_argument('--work-dir', '-w',
-                                 help='Directory used to store all temporary '
-                                 'files created during the process.')
+                                      'scripts',
+                                 # NOTE(maurosr): move this to /etc in the future
```
ghost: I believe this note could be removed.

lib/config.py
```diff
@@ -149,46 +49,61 @@ def __init__(self):
...
+                command_help = config_metadata[section]["help"]
+                command_parser = subparsers.add_parser(command, help=command_help)
+                target_parser = command_parser
+            else:
+                target_parser = self.parser
+            config_metadata[section].pop("help")
+            for attribute in config_metadata[section]:
+                argument = attribute.replace("_", "-")
+                argument_default = config_metadata[section][attribute]["default"]
+                argument_type = type(argument_default)
```
ghost: This will cause problems for arguments with `None` defaults.

lib/config.py
```diff
@@ -149,46 +49,61 @@ def __init__(self):
...
+            if section != "common":
+                command = section.replace("_", "-")
+                command_help = config_metadata[section]["help"]
+                command_parser = subparsers.add_parser(command, help=command_help)
+                target_parser = command_parser
+            else:
+                target_parser = self.parser
+            config_metadata[section].pop("help")
+            for attribute in config_metadata[section]:
+                argument = attribute.replace("_", "-")
```
ghost: Should we keep the hyphens instead of underscores in command line? This can be confusing when comparing with the config file.

lib/config.py
```diff
@@ -149,46 +49,61 @@ def __init__(self):
...
+                argument = "--%s" % argument
+                keyword_args = {"help": argument_help}
+                if argument_type is bool:
+                    if bool(argument_default):
+                        argument = "--no-%s" % argument[2:]
+                        action = "store_false"
+                    else:
+                        action= "store_false"
+                elif argument_type is list:
+                        keyword_args["nargs"] = "*"
```
ghost: Wrong indentation.

lib/config.py
```diff
@@ -149,46 +49,61 @@ def __init__(self):
...
+                argument = attribute.replace("_", "-")
+                argument_default = config_metadata[section][attribute]["default"]
+                argument_type = type(argument_default)
+                argument_help = config_metadata[section][attribute]["help"]
+                action = "store"
+                argument = "--%s" % argument
+                keyword_args = {"help": argument_help}
+                if argument_type is bool:
+                    if bool(argument_default):
+                        argument = "--no-%s" % argument[2:]
```
ghost: This requires setting the `dest` option to `attribute`

lib/config.py
ghost: I agree with @fabianorosas that setting all values in the config_metadata file could make the code less error prone and the configuration more flexible. This flexibility could be useful in the future. I do think that defaults could be left out to avoid making the file too big, however. The argparse defaults could be set like this, for example: http://stackoverflow.com/a/12627202

ghost (changes requested):

mopsfelder:
@alanoe As we discussed, we should avoid stacked pull requests as possible as we can.  At least, can you please state in the description which pull request this depends on?  For example:

```
Depends on #N
```

mopsfelder (changes requested):
Please update pull request description to properly track of dependencies.

alanoe:
MO> As we discussed, we should avoid stacked pull requests as possible as we can. At least, can you please state in the description which pull request this depends on?

This PR was created before we discussed that. Updating PR description.

lib/config.py
alanoe: Linux command line arguments usually use "-" as separator, but up to you guys.

config.yaml
alanoe: This is the default serialization behavior in PyYAML and yamllint does not complain.

lib/config.py
alanoe: I may do it in a separate patch, it is not related to this one.

lib/config.py

lib/config.py

lib/config.py
alanoe: No, it will not. Talked to Olav personally about this and he agreed

lib/config.py
```diff
@@ -149,46 +49,61 @@ def __init__(self):
...
+                        action = "store_false"
+                    else:
+                        action= "store_false"
```
ghost: "store_true"

alanoe:
LA> We need some documentation related to the config metadata file and the execution of sync_config_yaml.py.

Where this would go? CONTRIBUTING.md?


FR> My suggestion is to explicitly set these values in config_metadata now that this won't be edited by the user

This will make the config_metadata.yaml file unnecessarily *huge* ; I disagree, but whatever you guys prefer. However, this is not essential to this feature, so I'll put it in a separate PR.

alanoe:
start build

lib/config.py
```diff
@@ -66,11 +66,11 @@
 }
 MOCK_ARGS = {
     ('--mock-binary',):
-        dict(help='Mock binary path', default='/usr/bin/mock'),
+        dict(help='Mock binary path'),
```
ghost: You could mention in the commit longer message that those values are already in the YAML file.

lib/config.py
```diff
@@ -237,16 +237,20 @@ def parse_command_line_arguments(self, args):
         result = self.parser.parse_args(args)
         return vars(result)
 
-    def parse_command_line(self):
+    def parse_command_line(self, command_line_args):
```
ghost, line 154: This method is not used anymore, you could remove it or merge with `parse_command_line_arguments` to be able to pass `None`.

lib/config.py
```diff
@@ -267,57 +271,41 @@ def parse_config_file(self, config_file_path):
...
+        # update common config node with remaining command line arguments
+        node_name = "common"
+        for key, value in command_line_args.items():
+            if key in config[node_name]:
+                if value is not None:
                     config[node_name][key] = value
-                    args.pop(key)
+                command_line_args.pop(key)
+        # add subcommand to config
+        config["subcommand"] = command_line_args["subcommand"]
```
ghost: Shouldn't we raise an exception if there's anything left in `command_line_args`? Or `argparse` would have identified it already?

config_metadata.yaml
```diff
@@ -0,0 +1,219 @@
...
+  help: 'HTTP proxy URL'
+  default: ''
+ result_dir:
+  help: 'Directory used to store results'
+  default: 'result'
+  target_subcommands: ['build-packages', 'build-iso']
+ packages_metadata_repo_url:
+  help: 'Packages metadata git repository URL'
+  default: 'https://github.com/open-power-host-os/versions.git'
+  target_subcommands: ['build-packages', 'build-iso', 'build-release-notes', 'update-versions-readme', 'update-versions', 'update-metapackage']
```
ghost, line 31: This is not used in 'build-iso'

config_metadata.yaml
```diff
@@ -0,0 +1,219 @@
...
+ packages_metadata_repo_branch:
+  help: 'Packages metadata git repository branch'
+  default: 'master'
+  target_subcommands: ['build-packages', 'build-iso', 'build-release-notes', 'update-versions-readme', 'update-versions', 'update-metapackage']
```
ghost, line 35: This is not used in 'build-iso'

lib/config.py
```diff
@@ -178,50 +60,81 @@ class ConfigParser(object):
...
+                continue
+            command = section.replace("_", "-")
+            command_help = config_metadata[section]["help"]
+            config_metadata[section].pop("help")
+            _subparsers[command] = subparsers.add_parser(command, help=command_help,
+                                                         formatter_class=CustomHelpFormatter)
+            target_parser = _subparsers[command]
+            for attribute in config_metadata[section]:
+                if attribute == "help":
+                    continue
```
ghost, line 126: You don't need this, you've already popped "help" above.

lib/config.py
```diff
@@ -294,14 +295,22 @@ def parse(self, command_line_args=None):
...
         for key, value in command_line_args.items():
             if key in config[node_name]:
                 if value is not None:
-                    config[node_name][key] = value
+                    value_type = type(config[node_name][key])
+                    if value_type == str or value_type == bool or value_type == list:
+                        config[node_name][key] = value
+                    elif type(config[node_name][key]) == dict:
+                        config[node_name][key] = json.loads(value)
                 command_line_args.pop(key)
```
mopsfelder: How integer values, like `log_size`, would be handled?

What if we abstracted this assignment logic to avoid code duplication?  For example:

```python
def assign_config_value(name, value):
  value_type = type(name)
  if value_type == dict:
    name = json.loads(value)
  else:
    name = value
```

Then `assign_config_value` could be used:

```python
for key, value in command_line_args.items():
  if key in config[node_name]:
    if value is not None:
      assign_config_value(config[node_name][key], value)
    command_line_args.pop(key)
```

@olavphibm @fabianorosas What do you think?

lib/build_manager.py
```diff
@@ -44,8 +45,11 @@ def _build_packages(self, distro, packages):
...
         # create package builder based on distro
         if distro.name == "CentOS":
-            mock_config_file = CONF.get('build_packages').get(
-                'mock_config').get(distro.name).get(distro.version)
+            mock_config_file = os.path.join(
+                "mock_configs", distro.name, distro.version,
+                "%s-%s-%s.cfg" % (distro.name, distro.version, distro.architecture))
+            if not os.path.isfile(mock_config_file):
+                raise exception.BaseException("Mock config file not found at %s" % mock_config_file)
             package_builder = mockbuilder.Mock(mock_config_file)
```
mopsfelder, line 53: This is pinning the configuration file path to a value user cannot change, making it difficult for user to build using different configuration files.

👎

lib/iso_builder.py
```diff
@@ -46,8 +46,12 @@ def _run_mock_command(self, cmd):
...
             self.common_config.get('distro_version'),
             self.common_config.get('architecture'))
-        mock_config_file = self.config.get('mock_config').get(distro.name).get(
-            distro.version)
+        mock_config_file = os.path.join(
+            "mock_configs", distro.lsb_name, distro.version,
+            "build-iso-%s-%s-%s.cfg" % (distro.name, distro.version, distro.architecture))
+        if not os.path.isfile(mock_config_file):
+            raise exception.BaseException("Mock config file not found at %s" % mock_config_file)
+
```
mopsfelder, line 54: Same here.

👎

lib/config.py
ghost: I agree in having the function, however it needs some modifications to actually modify the dict:
```python
def assign_config_value(config_dict, key, value):
  value_type = type(config_dict[key])
  if value_type == dict:
    config_dict[key] = json.loads(value)
  else:
    config_dict[key] = value
```

olavph:
Superseded by https://github.com/open-power-host-os/builds/pull/251

PR #197, opened 2017-02-22
Refactor config - sync options in config file and command line

alanoe:
- Make options names consistent
- Sync options in config file and command line
- Rename config attributes to make them more consistent


lisibann (changes requested):
These changes are not related to config refactoring (config file and command line). I suggest a separate PR.

alanoe:
LA> These changes are not related to config refactoring (config file and command line). I suggest a separate PR.

This will cause a big rebase effort, I vote against it. Why can't you just review each patch here? If the issue is that the PR description is incomplete, I can simply complete it.

farosas:
start build

lib/config.py
```diff
@@ -263,9 +264,11 @@ def parse(self):
 
         # update iso node with iso subcommand args and then drop them from args
         for key, value in args.items():
-            if key in config['iso']:
-                config['iso'][key] = value
-                args.pop(key)
+            for node in COMMAND_TO_CONFIG_NODE.values():
```
ghost: Why do you need this for here?
Why not just `if key in config[COMMAND_TO_CONFIG_NODE[command_line_args.subcommand]]:`

lib/config.py
```diff
@@ -241,7 +241,7 @@ def parse(self):
         # Each subcommand may have a node for specific configurations
         # at the same level of the 'default' node
         COMMAND_TO_CONFIG_NODE = {
-            "build-iso": "iso",
+            "build-iso": "build_iso",
             "release-notes": "release_notes"
         }
```
ghost: Why not make this a list already?

lib/packages_manager.py
```diff
@@ -28,6 +30,7 @@ def __init__(self, packages_names):
         self.packages_names = packages_names
         self.packages = []
 
+
```
ghost: There's an extra line being added here, not following PEP8.

tests/unit/config_tests.py
```diff
@@ -38,20 +38,20 @@ def test_parse_arguments_list_WithLongArgument_ShouldParseArgumentValue(self, ar
...
+        (['build-packages'], 'packages', None),
+        (['build-packages'], 'result_dir', 'result'),
+        (['build-packages'], 'packages_metadata_repo_url', None),
+        (['build-packages'], 'packages_metadata_repo_branch', None),
+        (['build-packages'], 'mock_args', ''),
+        (['build-release-notes'], 'push_repo_url', None),
+        (['build-release-notes'], 'push_repo_branch', 'master'),
+        (['build-release-notes'], 'updater_name', None),
+        (['build-release-notes'], 'updater_email', None),
+        (['build-iso'], 'packages_dir', './result'),
```
ghost: This changed with gyr's patches. It should be `result/packages/latest`.

tools/upgrade_versions.py
```diff
@@ -225,7 +225,7 @@ def push_packages_head_commit(
 
 def run(CONF):
     versions_repo = setup_versions_repository(CONF)
-    packages_to_update = CONF.get('common').get('packages') or PACKAGES
+    packages_to_update = PACKAGES
```
ghost: Please do not remove this. The `upgrade-versions` command requires downloading the packages repositories to check if the version file has been updated. This takes a long time in a fresh development environment, so it's useful to execute with a limited set of packages (just a small one, like libvpd, for example).
The `build-release-notes` command do not download packages repositories, so this is not an issue above.

config.yaml
```diff
@@ -19,6 +19,10 @@ common:
    --plugin-option=tmpfs:required_ram_mb=39800"
  commit_updates: True
  push_updates: True
+build_packages:
+ packages:
+ keep_build_dir: False
+ result_dir:  "./result"
```
ghost: This is not needed. `result_dir` is already in the `common` section.

lib/iso_spinner.py
```diff
@@ -29,21 +29,26 @@
...
-        self.work_dir = config.get('common').get('work_dir')
-        self.timestamp = datetime.datetime.now().isoformat()
-        self.result_dir = os.path.join(config.get('common').get('result_dir'),
-                                       'iso', self.timestamp)
+        self.common_config = config.get('common')
         self.config = config.get("build_iso")
+        self.work_dir = self.common_config.get('work_dir')
+        self.timestamp = datetime.datetime.now().isoformat()
+        self.result_dir = os.path.join(self.common_config.get('result_dir'),
+            'iso', self.timestamp)
```
ghost: Wrong indentation.

lib/config.py
```diff
@@ -83,16 +83,35 @@
...
 }
+ISO_ARGS = {
+    ('--packages-dir', '-d'):
+        dict(help='Directory of packages used in the ISO image.',
+             default='result/packages/latest'),
+    ('--iso-name',):
+        dict(help='ISO name.',
+             default='OpenPOWER-Host_OS'),
+    ('--log-file',):
+        dict(help='ISO name.',
```
ghost: ISO log file path.

lib/config.py
```diff
@@ -83,16 +83,35 @@
...
+             default='/var/log/host-os/iso.log'),
+    ('--automated-install-file',):
+        dict(help='Path of a kickstart file, used to automate the installation of a RPM-based Linux distribution',
+             default='host-os.ks'),
+    ('--hostos-packages-groups',):
+        dict(help='Packages groups in yum repository'),
+    ('--automated-install-packages-groups',):
+        dict(help='Packages and packages groups which are installed using automated installation', nargs='*'),
+    ('--mock-iso-repo-name',):
+        dict(help='Name of the yum repository which will be created with OpenPOWER Host OS packages'),
```
ghost: I'd replace "which will be created" with "to create"

lib/config.py
```diff
@@ -83,16 +83,35 @@
...
+    ('--mock-iso-repo-dir',):
+        dict(help='Directory path of the yum repository which will be created with OpenPOWER Host OS packages'),
```
ghost: I'd replace "which will be created" with "to create"

lib/build_manager.py
```diff
@@ -15,11 +15,13 @@
 
 import logging
 
+import lib.centos
```
ghost: This should be after the last `from`

lib/iso_spinner.py
```diff
@@ -29,21 +29,26 @@
...
 
     def _run_mock_command(self, cmd):
+        distro = distro_utils.get_distro(
+            self.common_config.get('distro_name'),
+            self.common_config.get('distro_version'),
+            self.common_config.get('arch_and_endianness'))
         try:
             utils.run_command("%s -r %s %s %s" % (
-                self.mock_binary, self.config.get('mock_config'),
+                self.mock_binary, distro.config_file,
```
ghost: This attribute does not exist in `LinuxDistribution` class. It is just a local variable in `CentOS` constructor.

lib/iso_builder.py
```diff
@@ -46,9 +46,11 @@ def _run_mock_command(self, cmd):
             self.common_config.get('distro_name'),
             self.common_config.get('distro_version'),
             self.common_config.get('arch_and_endianness'))
+        mock_config_file = self.config.get('mock_config').get(distro.lsb_name).get(
+            distro.version)
```
ghost: This should probably be in the previous patch "Make ISO creation get distro version specific Mock config file "

tests/unit/config_tests.py
```diff
@@ -51,7 +51,7 @@ def test_parse_arguments_list_WithLongArgument_ShouldParseArgumentValue(self, ar
         (['build-release-notes'], 'push_repo_branch', 'master'),
         (['build-release-notes'], 'updater_name', None),
         (['build-release-notes'], 'updater_email', None),
-        (['build-iso'], 'packages_dir', './result'),
```
ghost: This was mistakenly changed in a previous patch of this PR.

ghost (changes requested):

lib/build_manager.py
alanoe: Done

lib/packages_manager.py
alanoe: Done

lib/config.py

lib/config.py

lib/config.py

lib/config.py
alanoe: Both will produce the same result. You are right, the for is unnecessary. Done

lib/config.py
alanoe: This dict will be deleted in a future patch, I prefer to not touch it here.

tools/upgrade_versions.py
alanoe: Not sure I follow. 

1) What is the purpose of testing with only a subset? It may work with the subset and break with a specific package outside the subset, we should test with all packages.
2) Do you have a subset of packages (Git/SVN) repositories in your dev environment? I have all in mine so I can test anything I want in the build system.

config.yaml
alanoe: Already fixed

lib/iso_spinner.py
alanoe: Done

lib/iso_spinner.py
alanoe: Done

config.yaml
```diff
@@ -22,6 +20,9 @@ default:
  commit_updates: True
  push_updates: True
  force_rebuild: False
+release_notes:
+ release_notes_repo_url: "https://github.com/open-power-host-os/open-power-host-os.github.io.git"
+ release_notes_repo_branch: "master"
```
mopsfelder: Not related to this change but do we need to repeat the name of the root node in its elements?  For example, `relete_notes_repo_url` is inside `release_notes`, it could be just `repo_url`.  Something like this:

```yaml
release_notes:
  repo_url:
  repo_branch:
```

mopsfelder (commented):
+1 for f345dab6cfcf56d11734244cfec96242b8b88de4

lisibann:
I reviewed all commits related to this PR. The comment was related to "Move selection of package builder and its conf file to BuildManager". No problem to keep it as is.

tools/upgrade_versions.py
ghost: 1. It is faster. Not every test during development needs to be complete.
2. Sometimes I need to clean the environment to test a new feature, sometimes the paths change (gyr's latest patches), sometimes I just don't want to wait to download the last week's of kernels updates.

config.yaml
ghost: @alanoe is making all options here match the same name as in `config.py`. Having a `repo_url` command line option there would be ambiguous. I believe this can be improved in future PRs, when options will be properly separated among commands.

ghost (approved):

lisibann (approved):

PR #200, opened 2017-02-22
Support disabling update of SCM repositories before build

alanoe:
Review only latest commit

### DIFF
--- a/lib/config.py
+++ b/lib/config.py
@@ -37,6 +37,9 @@ PACKAGE_ARGS = {
              nargs='*'),
 }
 PACKAGE_BUILD_ARGS = {
+    ('--no-update-packages-repos-before-build',):
+        dict(help='Do not update code repositories before building',
+             action='store_false', dest='update_packages_repos_before_build'),
     ('--keep-build-dir',):
         dict(help='Keep build directory and its logs and artifacts.',
              action='store_true'),
--- a/lib/package.py
+++ b/lib/package.py
@@ -13,7 +13,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from functools import partial
 from functools import total_ordering
 import fcntl
 import logging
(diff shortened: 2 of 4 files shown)


lisibann (approved):

lib/config.py
```diff
@@ -43,6 +43,9 @@
     ('--packages-repos-target-path', '-R'):
         dict(help='Directory where to clone code repositories',
              default='/var/lib/host-os/repositories'),
+    ('--no-update-packages-repos-before-build',):
+        dict(help='Update code repositories before building',
```
ghost: This option actually does the opposite: "Do not update code repositories before building"

ghost (changes requested):

olavphibm:
start build

alanoe:
start build

lib/package.py
```diff
@@ -118,9 +117,11 @@ def download_files(self, recurse=True):
         Download package source code and build files.
         Optionally, do the same for its dependencies, recursively.
         """
-        # Download all package sources
-        download_f = partial(package_source.download, directory=PACKAGES_REPOS_TARGET_PATH, local_copy_subdir_name=self.name)
-        self.sources = map(download_f, self.sources)
+        update_packages_repos = CONF.get('build_packages').get('update_packages_repos_before_build')
+        for source in self.sources:
+            if ['url'] not in source.keys() and update_packages_repos:
```
farosas: Why the exclusion of sources with 'url' in them?

alanoe: The file downloaded from a URL source is not cached in our build system, so it must always be downloaded. The logic was wrong, I fixed it.

lib/config.py
```diff
@@ -35,8 +35,9 @@
     ('--packages', '-p'):
         dict(help='Packages to be built',
              nargs='*'),
-}
-PACKAGE_BUILD_ARGS = {
```
ghost: This shouldn't be removed.

lib/config.py
```diff
@@ -116,7 +117,7 @@
 }
 SUBCOMMANDS = [
     ('build-packages', 'Build packages.',
-        [PACKAGE_ARGS, PACKAGE_BUILD_ARGS, MOCK_ARGS, DISTRO_ARGS, BUILD_REPO_ARGS, BUILD_ARGS]),
```
ghost: This shouldn't be removed.

ghost (changes requested):
The lack of of a GitRepository at `source['git']['repo']` will make _git_archive break. Maybe we should create `GitRepository` from the local directory anyway?

olavphibm:
start build

ghost:
start build

lib/package_source.py
```diff
@@ -10,6 +10,26 @@
 from lib import utils
 
 
+def set_dest_dir(source, directory, local_copy_subdir_name):
```
ghost, line 13: Could you document this method and mention the git repository is created from a local directory if it does not exist?

lib/package_source.py
```diff
@@ -10,6 +10,26 @@
...
+            repo_name = os.path.basename(source[source_type]['src'])
+            dest = os.path.join(directory, repo_name)
+    else:
+        file_name = os.path.basename(source['url']['src'])
+        dest = os.path.join(directory, file_name)
+
+    source[source_type]['dest'] = dest
+
+    # in git, git repo object must be created for later git archive
+    if source_type == 'git':
```
ghost: It's probably better to avoid re-creating the `GitRepository`:
`if source_type == 'git' and not source['git']['repo']:`

lib/package_source.py
alanoe: Done

lib/package_source.py
alanoe: - It seems to me that, currently, this will be only run once for each package in each host_os.py script execution
- If for some reason, it starts to run more than once, the 'src' attribute value in source may have changed since last execution and IMHO the Git repository object must be recreated

ghost: It is created in the `_git_download` function. The 'src' attribute is not modified anywhere.

alanoe: I understood your point now. I refactored code, having this code inside this function deviates from its scope and adding this if you suggested here is error prone, in case 'src' may change in the future.

ghost (approved):

farosas (approved):

PR #201, opened 2017-02-23
Make code bw compatible with old yaml files with incorrect dependencies

alanoe:


ghost (approved):

lib/rpm_package.py
```diff
@@ -205,6 +205,11 @@ def _load(self):
                     dep_name, self.distro, force_rebuild=self.force_rebuild)
                 self.build_dependencies.append(dep)
 
+            # keeps backward compatibility with old yaml files which have build dependencies
+            # listed as install dependencies
+            if self.distro == "7.2":
```
mopsfelder: Wouldn't it be `self.distro.version`?

alanoe: Yes, ty. Done

lib/rpm_package.py
```diff
@@ -205,6 +205,11 @@ def _load(self):
                     dep_name, self.distro, force_rebuild=self.force_rebuild)
                 self.build_dependencies.append(dep)
 
+            # keeps backward compatibility with old yaml files which have build dependencies
+            # listed as install dependencies
+            if self.distro.version == "7.2":
+                self.build_dependencies = list(set(self.build_dependencies + self.install_dependencies))
```
mopsfelder: This line has more than 80 columns.

lib/rpm_package.py
```diff
@@ -205,6 +205,11 @@ def _load(self):
                     dep_name, self.distro, force_rebuild=self.force_rebuild)
                 self.build_dependencies.append(dep)
 
+            # keeps backward compatibility with old yaml files which have build dependencies
```
mopsfelder: This line has more than 80 columns.

mopsfelder (changes requested):

lib/rpm_package.py
alanoe: Several lines in our code (including this file) have more than 80 columns and we are not reviewing this exact line limit in *any* of our code reviews. Pls let us be consistent in code review. This is as minor as it gets. I'll alter it here anyway.

mopsfelder: I understand that some code slipped into the tree without complying with PEP 8.  Perhaps that's a sign that we need to improve our automated checks and code validations.

If we have the opportunity to comply with PEP 8 during review, it's better for us.  This will result in less work in the future and code will get fixed gradually.

ghost (approved):

mopsfelder (approved):

PR #203, opened 2017-03-02
Create script to update Host OS metapackage

ghost:


ghost:
This depends on https://github.com/open-power-host-os/versions/pull/135

lib/versions_repository.py
```diff
@@ -61,29 +61,35 @@ def create_html_table(packages):
...
 
     Args:
-        file_path(str): file path
-        html_table (str): string representation of the HTML table which will
-            replace existing ones
+        file_path (str): file path
+        new_contents (str): contents that will replace the old ones
+        start_delimiter (str): delimiter marking start of contents to be
+            replaced
+        start_delimiter (str): delimiter marking end of contents to be
```
alanoe: start_delimiter => end_delimiter

tools/update_metapackage.py
```diff
@@ -0,0 +1,63 @@
...
+from lib.versions_repository import update_metapackage
+
+
+LOG = logging.getLogger(__name__)
+
+
+def run(CONF):
+    versions_repo = setup_versions_repository(CONF)
+    package_names = (CONF.get('default').get('packages') or
+                     config.discover_packages())
```
alanoe: When #197 is merged, you'll need to add this to a new update_metapackage node in config.yaml and put this 'packages' attribute under it to maintain consistency. Same for push_repo_url and push_repo_branch attributes below

lib/versions_repository.py
```diff
@@ -138,3 +139,118 @@ def read_version_and_milestone(versions_repo):
...
+
+    metapackage = rpm_package.RPM_Package.get_instance(
+        metapackage_name, distro)
+    metapackage.spec_file.update_version(metapackage_version)
+    metapackage.spec_file.replace_macro_definition(
+        "milestone", metapackage_milestone)
+
+    replace_spec_dependencies(metapackage.spec_file.path)
+    metapackage.spec_file.bump_release(
+        ["Update package dependencies"], user_name, user_email)
```
alanoe: IMHO these metapackage rpm spec handling methods should not be in versions_repository.py. As you want to reuse them, I suggest you create a new Python module.

alanoe (changes requested):

lib/versions_repository.py
ghost: Done.

tools/update_metapackage.py
ghost: Removed the option and added only push_repo_* to yaml.

lib/versions_repository.py
ghost: Created `metapackage` module

PR #207, opened 2017-03-07
Support overriding package source code repositories data in command line

alanoe:


lisibann (approved):

lib/package.py
```diff
@@ -101,8 +101,31 @@ def __init__(self, name, force_rebuild=True):
...
         self._load()
 
+        # get global config information which may override YAML
+        packages = CONF.get('build_packages').get('packages') or []
+        for package in packages:
+            if package.startswith(name):
+                if package == name:
+                    break
+                else:
+                    package_parts = package.split("#")
```
mopsfelder: I think adding a comment here describing the expected format helps.  For example:

```python
# Parse and set package source attributes from the command line.
# Fields are separated by `#` character and the expected format
# can be one of the following:
#
#     package_name#repo_url#branch_name#revision_id
#     package_name#repo_url#reference
#     package_name##reference
#     package_name

package_parts = package.split("#")
...
```

ghost: I think this format should also be in the help of the "packages" option.

lib/package.py
```diff
@@ -101,8 +101,31 @@ def __init__(self, name, force_rebuild=True):
             raise exception.PackageDescriptorError(
                 "Failed to find %s's YAML descriptor" % self.name)
 
+        # load package metadata YAML file
         self._load()
 
+        # get global config information which may override YAML
+        packages = CONF.get('build_packages').get('packages') or []
+        for package in packages:
+            if package.startswith(name):
```
ghost: This could cause a problem with ginger and ginger-base, for example. You should split the parts before this check.

tools/build_packages.py
```diff
@@ -32,6 +32,11 @@ def run(CONF):
         CONF.get('common').get('distro_version'),
         CONF.get('common').get('arch_and_endianness'))
 
-    LOG.info("Building packages: %s", ", ".join(packages_to_build))
-    bm = build_manager.BuildManager(packages_to_build, distro)
+    # get packages names
+    packages_to_build_names = []
+    for package in packages_to_build:
+        packages_to_build_names.append(package.split("#")[0])
```
ghost, line 38: This must also be done for update-versions, unless we want to break consistency of the command line option and mention the differences in the help message.

lib/package.py
```diff
@@ -101,8 +101,31 @@ def __init__(self, name, force_rebuild=True):
...
+                    if package_parts[1]:
+                        self.sources[0][source_type]["src"] = package_parts[1]
+                    if len(package_parts) == 4:
+                        if package_parts[2]:
+                           self.sources[0][source_type]["branch"] = package_parts[2]
+                        if package_parts[3]:
+                           self.sources[0][source_type]["commit_id"] = package_parts[3]
+                    elif len(package_parts) == 3:
+                        if package_parts[2]:
+                           self.sources[0][source_type]["commit_id"] = package_parts[2]
```
ghost: I find it a little hard to follow the code with the nested ifs and indexes everywhere.
I know you're not a fan of EAFP, but here's a suggestion. The try block stops when it finds an undefined variable. Feel free to keep as is if you dislike it.
```
package_parts = package.split("#")
if len(package_parts) == 1:
    break
elif len(package_parts) == 2:
    _, src = package_parts
elif len(package_parts) == 3:
    _, src, commit_id = package_parts
elif len(package_parts) == 4:
    _, src, branch, commit_id = package_parts
else:
    raise "Too many parts"
try:
    if src:
       self.sources[0][source_type]["src"] = src
    if commit_id:
       self.sources[0][source_type]["commit_id"] = commit_id
    if branch:
       self.sources[0][source_type]["branch"] = branch
except NameError:
    pass
```

ghost (changes requested):

lib/package.py

lib/package.py
alanoe: I indeed dislike EAFP, but I improved the code:
 - reduced if nesting a little
 - replaced self.sources[0][source_type] with a variable
 - tested if there are any sources

tools/build_packages.py
alanoe: Done

lib/package.py

olavphibm:
start build

lib/config.py
```diff
@@ -141,6 +148,14 @@ def get_config():
     return config_parser
 
 
+class CustomHelpFormatter(argparse.HelpFormatter):
+    def _split_lines(self, text, width):
+        if text.startswith('R|'):
```
ghost: Why `R|`?

lib/package.py
```diff
@@ -101,8 +101,46 @@ def __init__(self, name, force_rebuild=True):
...
+                    break
+                # assume that the first source is the main one and that a source
+                # override targets it
+                source_type = self.sources[0].keys()[0]
+                main_source = self.sources[0][source_type]
+                if len(package_parts) == 1:
+                    pass
+                elif len(package_parts) == 2:
+                    if package_parts[1]:
+                        main_source["src"] = package_parts[1]
```
ghost, line 130: `src` should also be updated when `package_parts` is 3 or 4.
I think you could remove this if `== 2` and use a `break` in `== 1`

ghost (changes requested):

lib/config.py
alanoe: A special "raw text" identifier, I need it to set that I want to use custom word wrap in the help msg. I updated the code to use a constant instead.

lib/package.py
alanoe: I still need to test if len == 2, if len >4 or something similar to be able to infer if number of parts is correct. IMHO testing each length is more readable.

olavphibm:
start tests

olavphibm:
start build

ghost (approved):

farosas (approved):

PR #211, opened 2017-03-09
Create yum repository files when building packages

ghost:


lib/mockbuilder.py
```diff
@@ -195,12 +203,32 @@ def copy_results(self, package):
...
+    def create_repository(self):
+        """
+        Create yum repository in build results directory.
+        """
+        result_dir = CONF.get('common').get('result_dir')
+        build_results_dir = os.path.join(
+            result_dir, 'packages', self.timestamp)
+        utils.run_command("createrepo %s" % build_results_dir)
+
+        repo_files_dir = os.path.join(result_dir, "repository_files")
```
mopsfelder: I don't like bike-shedding discussion but `repository_files` sounds awkward to me.

What if we shortened it to just `repository` then `result` directory would look like the following:

```
result/repository/
result/iso/
...
```

It's implicit that some *files* will be inside *repository* directory, we don't need to re-state that in the directory name.

We can still change it a little bit to something similar to what is done in [CentOS image](http://mirror.centos.org/altarch/7/os/ppc64le/):

```
result/packages/
result/repodata/
result/iso/
...
```

What do you think?

ghost: The actual repository is saved to `result/packages/<timestamp>`
The `result/repository_files` directory is supposed to keep only the `.repo` files, to ease the utilization of the repositories.
Another option would be to add another level to the former directory, i.e. `result/packages/<timestamp>/repository`, and putting the `.repo` file in `result/packages/<timestamp>`. But I prefer to keep them separated.

lib/mockbuilder.py
```diff
@@ -199,16 +207,28 @@ def create_repository(self):
         """
         Create yum repository in build results directory.
         """
+        result_dir = CONF.get('common').get('result_dir')
         build_results_dir = os.path.join(
-            CONF.get('common').get('result_dir'), 'packages', self.timestamp)
+            result_dir, 'packages', self.timestamp)
         utils.run_command("createrepo %s" % build_results_dir)
 
+        repo_files_dir = os.path.join(result_dir, "repository_files")
```
alanoe: - Why "files" in "repository_files" if there is only one file?
- Agree with Murilo that "files" do not add anything
- It seems you are assuming that "repository_files" == .repo file and "repository" == packages + repository metadata files automatically created by repodata. 

Could we make this clearer? I suggest repository_files => repository_config

lib/mockbuilder.py
```diff
@@ -195,6 +195,14 @@ def copy_results(self, package):
             self.timestamp, package.name)
         self._copy_rpms(package.build_cache_dir, package_build_results_dir)
 
+    def create_repository(self):
+        """
+        Create yum repository in build results directory.
+        """
+        build_results_dir = os.path.join(
+            CONF.get('common').get('result_dir'), 'packages', self.timestamp)
+        utils.run_command("createrepo %s" % build_results_dir)
```
alanoe: minor: YUM repository creation has nothing to do with Mock, IMHO this should be in a different Python module.

lib/mockbuilder.py
```diff
@@ -30,6 +30,14 @@
 CONF = config.get_config().CONF
```
alanoe, line 30: Regarding commit msg, it took me a while to understand that "yum repository files" here do not refer to the yum repository metadata files, but only to the yum repository config file. I'd prefer "Create yum repository config file when building packages"

lib/mockbuilder.py
mopsfelder: Perhaps the timestamp should be the father of the resulting descending directories of a build, for example:

```
result/<timestamp>/packages/<file>.rpm
result/<timestamp>/repository/<file>.repo
result/<timestamp>/repository/repodata/<file>.xml
result/<timestamp>/iso/<file>.iso
```

This facilitates if we want to move the output of a build to somewhere, i.e. we would just need to move the `result/<timestamp>` directory and all output will be together, no need to pickup sparse directories in the filesystem.

README.rst
```diff
@@ -128,60 +128,31 @@ file or disable package repositories update in config.yaml.
 Using the RPMs
 --------------
 
-We decided that's is a little bit intrusive to install all the
-produced RPMs since some of them may not fit your needs. At the end of
```
alanoe: Should we remove this first sentence?

README.rst
```diff
@@ -128,60 +128,31 @@ file or disable package repositories update in config.yaml.
...
-the process you'll see a new ``result`` directory inside this
-project's root.
+At the end of the process you'll see a new ``result`` directory inside
+this project's root. It contains the built packages repositories under
+``packages`` and yum repository files under ``repository_files`` to
```
alanoe: yum repository files => a yum repository config file

to avoid ambiguity

Also, repository_files => repository_config

README.rst
```diff
@@ -128,60 +128,31 @@ file or disable package repositories update in config.yaml.
...
+simplify using those repositories.
 
-A suggested set of packages tested is the following:
+To install only virtualization related packages, for example, you can
```
alanoe: - You are altering 2 things in the same patch: mentioning the new yum repository config file and mentioning the new metapackages. IMHO the latter should be in a separate patch.
- It'd be nice to mention the list of RPM packages installed by each metapackage in some doc (AFAIK it is not) and not force the user to read rpm .spec files to figure it out.

alanoe (changes requested):

lib/mockbuilder.py
alanoe: I like this ^ . It deviates from traditional yum repo organization, in which the repodata dir is in the same dir of the packages, but I dont know if that is a problem.

lib/mockbuilder.py
ghost: Done.

lib/mockbuilder.py
ghost: Renamed to repository_config

README.rst
ghost: Done.

README.rst
ghost: - Separated them.
- I agree. Those lists should be in the versions repository and it would be nice if they were updated automatically with the `update-versions` command.

lib/mockbuilder.py
ghost: I don't currently see the benefit of creating a module for a single function that will only be called from here.

README.rst
ghost: We are recommending only the virtualization packages and the user is executing the install command himself, so he has control over what's installed. I think this sentence is unnecessary.

olavphibm:
start build

lib/mockbuilder.py
alanoe: IMHO it makes more sense to put this in utils.py or a new yum_repository.py. You could put the yum repository config file creation there too in the later patch which does it. This module should contain *only* Mock-specific features. But up to you.

alanoe (approved):

lisibann (approved):

lib/mockbuilder.py
ghost: @mopsfelder I don't know if that's a good idea. With this structure, we wouldn't know if `result/latest` points to a `builds-packages` result or a `build-iso` result. The `repository_config` files that I'm creating here are also only useful in the system in which the build was executed. It would make no sense to move it around since the paths have to be hardcoded in yum config.

farosas (approved):

PR #212, opened 2017-03-10
Fix mock incomplete output to log file

gyribeiro:


config.yaml
```diff
@@ -39,6 +39,12 @@ build_iso:
      CentOS:
       '7': "./mock_configs/CentOS/7/build-iso-CentOS-7-ppc64le.cfg"
     mock_args:
+      # If the system has less RAM than required, the plugin will be disabled
+      "--enable-plugin=tmpfs
+      --plugin-option=tmpfs:keep_mounted=True
+      --plugin-option=tmpfs:max_fs_size=32g
+      --plugin-option=tmpfs:required_ram_mb=39800
```
ghost, line 47: Do you any idea how much disk space and RAM the ISO creation uses? Those values could probably be reduced.

alanoe: - I assume this change aims performance increase, pls document this in commit longer msg
- Certainly 32G for max FS size and 39800 MB for RAM is too much for ISO creation

alanoe (changes requested):

farosas (approved):

mopsfelder:
Let's merge this and tune those numbers later.

Thanks, @gutoyr!

PR #220, opened 2017-03-16
Add Host OS package environments to ISO installation

ghost:


olavphibm:
build iso

olavphibm:
start iso

config.yaml
```diff
@@ -40,6 +40,8 @@ build_iso:
     mock_args:
```
alanoe: Regarding commit msg, pls explain explicitly that a "package environment" is a group of package groups.

config.yaml
```diff
@@ -40,6 +40,8 @@ build_iso:
...
     packages_dir: "result/packages/latest"
     automated_install_file: "host-os.ks"
+    base_packages_groups:
```
alanoe: I suggest "base_distro_minimal_packages_groups" or  base_distro_base_packages_groups" instead of "base_packages_groups". Also, document here:
- what an environment is 
- that this represents the base distro's minimal list of package groups which must be installed so that the OS works
- that these packages groups which be available in all package environments

lib/config.py
```diff
@@ -103,6 +103,9 @@
     ('--automated-install-file',):
         dict(help='Path of a kickstart file, used to automate the installation of a RPM-based Linux distribution',
              default='host-os.ks'),
+    ('--base-packages-groups',):
+        dict(help='Base packages groups that should always be installed in '
+             'the target system', nargs='*'),
```
alanoe: Base here means both "from base distro" and "minimal". Could you make this more explicit ?

lib/packages_groups_xml_creator.py
```diff
@@ -18,6 +18,20 @@
 
 COMPS_DOCTYPE = "<!DOCTYPE comps PUBLIC \"-//HostOS//DTD Comps info//EN\" \"comps.dtd\">"
 
+def convert_name_to_id(element_name, suffix):
```
alanoe: minor side comment: appending a suffix is out of the scope of "converting a name to ID", but as this feature is reused, OK

lib/packages_groups_xml_creator.py
```diff
@@ -67,7 +99,39 @@ def create_group_xml(group_name, group_pkgs):
     return grp
 
 
-def create_comps_xml(packages_groups):
+def create_environment_xml(environment_name, environment_groups):
+    """
+    Construct an XML object representing an 'environment' element.
```
alanoe, line 106: You could explain explicitly that an environment is simply a named list of packages groups somewhere.

lib/packages_groups_xml_creator.py
```diff
@@ -67,7 +99,39 @@ def create_group_xml(group_name, group_pkgs):
...
+    Those elements will appear as options during the installation.
+
+    Args:
+        environment_name (str): name of an environment of package groups
```
alanoe: If you explain what an  environment is above, IMHO you could simply use "packages environment" instead of "environment of packages groups" here.

config.yaml
```diff
@@ -41,34 +41,18 @@ build_iso:
...
+        OpenPOWER Host OS base:
+            - "open-power-host-os-base"
+        OpenPOWER Host OS container:
+            - "open-power-host-os-container"
+        OpenPOWER Host OS virtualization:
+            - "open-power-host-os-virt"
+        OpenPOWER Host OS virtualization management:
+            - "open-power-host-os-virt-management"
+        OpenPOWER Host OS RAS:
+            - "open-power-host-os-ras"
```
alanoe, line 63: if there will be a 1:1 relationship between package group and metapackages, couldn't we automate the creation of this list with code? You could keep the list of metapackages somewhere (or even better, assume they are the ones which start with open-power-host-os in release rpm spec in versions) and create the groups names from it.

alanoe (changes requested):

config.yaml
ghost: Mentioned in argparse's help

config.yaml
ghost: I prefer to leave this explicit and flexible here than buried inside the code.
There's also, currently, one package that is not here: open-power-host-os-release.

lib/packages_groups_xml_creator.py
ghost: Done

lib/packages_groups_xml_creator.py
ghost: Done

farosas (approved):

olavphibm:
start build

lib/config.py
```diff
@@ -117,8 +117,9 @@
              default='host-os.ks'),
     ('--host-os-packages-groups',):
         dict(help='Packages groups in yum repository'),
-    ('--automated-install-packages-groups',):
-        dict(help='Packages and packages groups which are installed using automated installation', nargs='*'),
+    ('--iso-root-fs-packages-groups',):
+        dict(help='Packages and packages groups that will be installed in '
+             'the ISO root file system.', nargs='*'),
```
alanoe: These packages and packages groups belong to the base distro (CentOS), it seems? It'd be good to mention it somewhere.

config.yaml
```diff
@@ -42,34 +42,18 @@ build_iso:
     packages_dir: "result/packages/latest"
```
alanoe, line 49: minor: It'd be good to mention somewhere in commit msg that you are not only *creating* new groups, but *replacing* the old packages groups.

config.yaml
```diff
@@ -41,7 +41,7 @@ build_iso:
     mock_args:
     packages_dir: "result/packages/latest"
     automated_install_file: "host-os.ks"
-    host_os_packages_groups:
+    installable_environments:
```
alanoe: why you removed "host_os" prefix? I'd prefer to differ the things which belong to Open Power Host OS and the ones which belong to Cent OS (base distro) eg the groups below

config.yaml
```diff
@@ -54,7 +54,9 @@ build_iso:
             - "open-power-host-os-virt-management"
         OpenPOWER Host OS RAS:
             - "open-power-host-os-ras"
-    iso_root_fs_packages_groups::
+    minimal_install_groups:
+        - "core"
```
alanoe: - I'd prefer to make it clear in var name this belongs to base distro (CentOS), hard to know these groups belongs to the base distro (Cent OS) and not OP Host OS without reading help below. Maybe you consider base_distro_minimal_install_groups is too long? Up to you.
- I'd prefer to refer to packages groups consistently. Why here core does not have a @ and below it does?

lib/packages_groups_xml_creator.py
```diff
@@ -82,21 +101,64 @@ def create_group_xml(group_name, group_pkgs):
...
-        packages_groups ({str:[str]}): dictionary where key is the package group
-                                       name and value is the list of packages
-                                       that are member of the group
+        packages_environments ({str: [str]}): dictionary where key is
+            the package environment and value is the list of packages
+            that will (indirectly) belong to that environment. An
+            intermediary packages group with the same name will be
+            created to contain those packages.
+        minimal_install_groups ([str]): packages groups that must be
+            present in every CentOS installation and will be added to
```
alanoe: we are hardcoding "CentOS" instead of using "base distro (eg CentOS)"?

olavphibm:
start iso

config.yaml
ghost: Done.

config.yaml
ghost: The existing CentOS packages groups are kept. I am not replacing them.

lib/packages_groups_xml_creator.py
ghost: Fixed.

lib/config.py
ghost: Mentioned that the groups passed to `--host-os-packages-groups` option are automatically added.

config.yaml
ghost: Those environments do not have to contain Host OS packages. They do have CentOS groups too, coming from the `minimal_install_groups` option.
If, for some reason, someone asks us to include an installation option to install a set of CentOS packages from extras, or whatever, this could be used.

config.yaml
ghost: - OK
- Splitted lists for groups and single packages (`grub2`) below to remove the confusing `@`s.

olavphibm:
start build

olavphibm:
start build

config.yaml
alanoe: You are deleting the previous OP Host OS virtualization-hypervisor package group...

config.yaml
alanoe: OK

config.yaml
ghost: Actually, this is a CentOS group that was modified by us, it will still exist.
Clarifying this in commit message.

alanoe (approved):

PR #223, opened 2017-03-20
Reduce changelog on spec files on update_versions

lisibann:
Currently, when we update packages versions/releases, we mention in the changelog every commit that is included, with its short commit message. This makes the changelog very big, especially for some packages such as kernel, with information that is not very useful by its own.

### DIFF
--- a/tools/update_versions.py
+++ b/tools/update_versions.py
@@ -51,22 +51,20 @@ PACKAGES = [
 PRERELEASE_TERMS = ['rc']
 
 
-def _get_git_log(repo, since_id):
+def _get_git_commit_log(repo, commit_id):
     """
-    Get log of commit SHA1 and short messages since the ID provided
-    (non-inclusive).
+    Get log of short SHA1 and short message for commit_id
     """
+    SHORT_HEX_SHA_LENGTH = 7
     log = []
-    for commit in repo.iter_commits():
-        if commit.hexsha.startswith(since_id):
-            break
-        commit_message = commit.message.split('\n')[0]
-        commit_message = commit_message.replace("'", "")
-        commit_message = commit_message.replace("\"", "")
-        log.append("%s %s" % (commit.hexsha, commit_message))
+    commit = repo.commit(commit_id)
+    commit_message = commit.message.split('\n')[0]
+    commit_message = commit_message.replace("'", "")
+    commit_message = commit_message.replace("\"", "")
+    log.append("Updating to %s %s" % (
+        commit.hexsha[0:SHORT_HEX_SHA_LENGTH], commit_message))
 
     return log
-
 
 class Version(object):
     def __init__(self, pkg):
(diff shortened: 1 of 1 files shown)


tools/update_versions.py
```diff
@@ -118,8 +132,8 @@ def _bump_release(self, pkg, change_log_header=None, user_name=None,
             new_commit_id = new_source["commit_id"]
             LOG.info("Updating package %s from %s to %s" % (
                 self.pkg.name, old_commit_id, new_commit_id))
-            change_log_lines += _get_git_log(
```
farosas, line 121: The old function is not used anywhere else, afaics. It should be removed.

tools/update_versions.py
```diff
@@ -118,8 +132,8 @@ def _bump_release(self, pkg, change_log_header=None, user_name=None,
...
-                new_source["repo"], old_commit_id)
+            change_log_lines += _get_git_last_commit_log(
+                new_source["repo"], new_commit_id)
```
ghost: You could use the [commit function](http://gitpython.readthedocs.io/en/1.0.1/reference.html#git.repo.base.Repo.commit) and [summary attribute](http://gitpython.readthedocs.io/en/1.0.1/reference.html#git.objects.commit.Commit.summary) to make this much simpler and remove the functions `_get_git_last_commit_log` and `_get_git_log`:
```
SHORT_HEX_SHA_LENGTH = 7
commit = new_source["repo"].commit(new_commit_id)
change_log_lines += "Updating to commit %s: %s" % (
    commit.hexsha[:SHORT_HEX_SHA_LENGTH], commit.summary))
```

Also, git usually only uses 7 character for short hashes.

ghost (changes requested):

tools/update_versions.py
```diff
@@ -67,6 +67,20 @@ def _get_git_log(repo, since_id):
 
     return log
 
+def _get_git_last_commit_log(repo, last_id):
+    """
+    Get log of short SHA1 and short message for last commit
+    """
+    log = []
+    for commit in repo.iter_commits():
```
farosas: Use `repo.commit` instead of iterating.

https://gitpython.readthedocs.io/en/stable/tutorial.html#the-commit-object

tools/update_versions.py
```diff
@@ -67,6 +67,20 @@ def _get_git_log(repo, since_id):
 
     return log
 
+def _get_git_last_commit_log(repo, last_id):
```
farosas: I think we could make this function generic. It doesn't need to know that it is using the last commit. Maybe `_get_git_log` or `_get_git_commit_log` and then alter the `last_id` argument to `commit_id`.

tools/update_versions.py
```diff
@@ -67,6 +67,20 @@ def _get_git_log(repo, since_id):
...
+    """
+    Get log of short SHA1 and short message for last commit
```
farosas: s/ last//

farosas (changes requested):

tools/update_versions.py
farosas: @olavphibm what about the removal of quotes?

ghost: @fabianorosas Ok, that may be a reason to keep the function. Why do we need to remove quotes though?

farosas: rpmdev-bumpspec going crazy maybe?

tools/update_versions.py
lisibann: done.

tools/update_versions.py
lisibann: done.

tools/update_versions.py
lisibann: done.

tools/update_versions.py
lisibann: done

farosas (approved):

tools/update_versions.py
lisibann: partially done.

ghost (approved):

PR #224, opened 2017-03-21
Define commit ID macros via rpmbuild's command line

ghost:
This allows us to remove the duplicated commit ID from the spec files.

### DIFF
--- a/lib/mockbuilder.py
+++ b/lib/mockbuilder.py
@@ -69,8 +69,10 @@ class Mock(package_builder.PackageBuilder):
         LOG.info("%s: Starting build process" % package.name)
         self._build_srpm(package)
         self._install_external_dependencies(package)
-        cmd = (self.common_mock_args + " --rebuild %s --no-clean --resultdir=%s"
-               % (self.build_dir + "/*.rpm", self.build_dir))
+        cmd = (self.common_mock_args
+               + " --rebuild %s --no-clean --resultdir=%s %s" % (
+                   self.build_dir + "/*.rpm", self.build_dir,
+                   package.get_spec_macros()))
 
         if package.rpmmacro:
             cmd = cmd + " --macro-file=%s" % package.rpmmacro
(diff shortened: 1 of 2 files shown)


lib/mockbuilder.py
```diff
@@ -69,8 +69,10 @@ def build(self, package):
         LOG.info("%s: Starting build process" % package.name)
         self._build_srpm(package)
         self._install_external_dependencies(package)
-        cmd = (self.common_mock_args + " --rebuild %s --no-clean --resultdir=%s"
-               % (self.build_dir + "/*.rpm", self.build_dir))
+        cmd = (self.common_mock_args
+               + " --rebuild %s --no-clean --resultdir=%s" % (
+                   self.build_dir + "/*.rpm", self.build_dir)
+               + package.get_spec_macros())
```
alanoe: minor: I do not understand why mix %s and + . Why not "--rebuild %s --no-clean --resultdir=%s %s" %s (self.build_dir + "/*.rpm", self.build_dir, package.get_spec_macros()) ?

lib/rpm_package.py
```diff
@@ -255,6 +259,35 @@ def _load(self):
...
+        """
+        macros_string = ""
+        for package_attribute, macro_name in (
+                PACKAGE_METADATA_TO_RPM_MACRO_MAPPING.items()):
+            subnode = self.package_data
+            for subnode_id in package_attribute:
+                try:
+                    subnode = subnode[subnode_id]
+                except (KeyError, IndexError):
+                    LOG.debug(
```
alanoe: This logs a debug msg for each mappings which is not found. IMHO it should log an error msg (and possibly cancel the build instead of failing silently) if none of the mappings are found.

lib/rpm_package.py
```diff
@@ -28,6 +28,10 @@
 
 CONF = config.get_config().CONF
 LOG = logging.getLogger(__name__)
+PACKAGE_METADATA_TO_RPM_MACRO_MAPPING = {
+    ("sources", 0, "git", "commit_id"): "git_commit_id",
+    ("sources", 0, "svn", "commit_id"): "svn_revision",
+}
```
alanoe: - I'd document that we are assuming that the first source is the main one
- Shouldn't we support Mercurial too?

alanoe (changes requested):

lib/mockbuilder.py
ghost: It is already mixed by having the `self.common_mock_args +`, but ok.

lib/rpm_package.py
ghost: - Done.
- Since, we have code to support it, it makes sense to be consistent.

lib/rpm_package.py
ghost: Some packages do not have commit IDs in their spec files because they correspond exactly to the version's branch in their repository (e.g. golang, systemtap, libnl3).
We also have the `open-power-host-os` package that do not have external sources and, consequently, would fail with no mappings.
I think it's enough to print a message in debug level if no mappings are found.
I am also logging a message when the mapping is found instead of in this `except` block, to avoid too many messages. The mapping not found can be inferred.

alanoe (approved):

PR #225, opened 2017-03-22
Refactor LinuxDistribution class

ghost:
Improve logs, warnings, variable names and documentation.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -8,7 +8,7 @@ common:
  http_proxy: ''
  distro_name: 'CentOS'
  distro_version: '7'
- arch_and_endianness: 'ppc64le'
+ architecture: 'ppc64le'
  mock_binary: "/usr/bin/mock"
  commit_updates: True
  push_updates: True
--- a/lib/build_manager.py
+++ b/lib/build_manager.py
@@ -43,9 +43,9 @@ class BuildManager(object):
         """
 
         # create package builder based on distro
-        if distro.lsb_name == "CentOS":
-            mock_config_file = CONF.get('build_packages').get('mock_config').get(
-                distro.lsb_name).get(distro.version)
+        if distro.name == "CentOS":
+            mock_config_file = CONF.get('build_packages').get(
+                'mock_config').get(distro.name).get(distro.version)
             package_builder = mockbuilder.Mock(mock_config_file)
         else:
             raise exception.DistributionError()
--- a/lib/distro.py
+++ b/lib/distro.py
@@ -21,22 +21,31 @@ from lib import exception
 LOG = logging.getLogger(__name__)
 # NOTE(maurosr): make it a constant since we only plan to work with little
 # endian GNU/Linux distributions.
-SUPPORTED_ARCH_AND_ENDIANNESS = ("PPC64LE")
+SUPPORTED_ARCHITECTURES = ("PPC64LE")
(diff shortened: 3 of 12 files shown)


config.yaml
```diff
@@ -8,7 +8,7 @@ common:
  http_proxy: ''
```
alanoe, line 8: I assume that we do not parse the endianess from the arch name and thus we do not require the endianess anywhere in our code? Otherwise, it would not support an architecture name without endianess (eg 'ppc64') and the old name would be better.

lib/distro.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, name=None, version=None, architecture=None):
                 distribution
         """
         self.lsb_name = name
-        if architecture.upper() not in SUPPORTED_ARCHITECTURES:
+        if architecture.lower() not in SUPPORTED_ARCHITECTURES:
```
alanoe: Using lower or upper case is a random personal choice, both work. Thus, I see no problem in this.

lib/build_manager.py
```diff
@@ -43,9 +43,9 @@ def _build_packages(self, distro, packages):
         """
```
alanoe, line 43: I disagree that 'lsb' does not add useful information. It means 'Linux Standard Base' and it is a Linux distribution unique ID. You could rename to lsb_id if you prefer.

"Name" is not a standardized value. Why we'd use "CentOS" instead of "Cent OS", for instance?

alanoe (changes requested):
See prev comments

lib/build_manager.py
alanoe: Never mind my previous comment. I just noticed we do not obtain lsb_name from the system installed with the Linux distribution but from a value provided by user (via config file or command line). And this is compared against a list of constants we define.

alanoe (approved):

lib/iso_builder.py
```diff
@@ -45,8 +45,8 @@ def _run_mock_command(self, cmd):
         distro = distro_utils.get_distro(
             self.common_config.get('distro_name'),
             self.common_config.get('distro_version'),
-            self.common_config.get('arch_and_endianness'))
-        mock_config_file = self.config.get('mock_config').get(distro.lsb_name).get(
+            self.common_config.get('architecture'))
```
farosas, line 48: Wrong commit

farosas (approved):

lib/distro.py
ghost: Not so random. Googling `ppc64le` returns only lower case matches in the first pages and the `platform.machine` function, as well as `uname -m` also return it lowercase.
We were also setting `arch_and_endianness: 'ppc64le'` in YAML, so it was inconsistent.

lisibann (approved):

config.yaml
ghost: That's right. We just treat it as a single string.

lib/iso_builder.py
ghost: Fixed.

lib/build_manager.py
ghost: I was not able to find a list of unique IDs defined by the Linux Standard Base, so I don't think that's the case. That's why I decided to drop this.

PR #227, opened 2017-03-23
Fix update of release tags in Host OS metapackage

ghost:


lib/rpm_package.py
```diff
@@ -72,26 +72,31 @@ def write_content(self):
...
+        else:
+            command = "rpmspec --srpm -q --qf '%%{%s}' %s %s 2>/dev/null" % (
+                    tag_name.upper(), self.path, extra_args)
+            tag_value = utils.run_command(command).strip()
             if tag_value == "(none)":
                 tag_value = None
-            self._cached_tags[tag_name] = tag_value
+            if not extra_args:
+                # Extra arguments may define macros that alter the tags
+                self._cached_tags[tag_name] = tag_value
```
farosas: Instead of bypassing the cache I suggest you keep the old code structure and move this block up with:
```
# reason why we need to invalidate the cache entry here.
if extra_args:
    del self._cached_tags[tag_name]

if tag_name not in self._cached_tags:
(...)
```

That way there's no special case in cache handling and I think it's easier to understand.

ghost: This way, if we call `query_tag` with `extra_args` and later without it, we might have an unexpected result.
This could happen if we wanted to overwrite a value that is present in the spec file, for example.

farosas (approved):

PR #234, opened 2017-03-29
Remove backwards compatibility with versions git repository older commits

alanoe:


ghost (changes requested):
We are not testing backwards compatibility and do not have a "stable" commit to build old versions yet, so we are not prepared to remove it right now.

alanoe:
OH> We are not testing backwards compatibility and do not have a "stable" commit to build old versions yet, so we are not prepared to remove it right now.

Indeed. I was not aware that the Jenkins job which build old host os versions was failing. I'll fix it, then we can do this.

farosas:
@alanoe please add some text explaining why we are removing backwards compatibility

PR #236, opened 2017-03-30
Move RPM creation to separate method

alanoe:


lib/mockbuilder.py
```diff
@@ -114,13 +132,25 @@ def _build_rpm(self, package):
             raise
 
     def prepare_sources(self, package):
+        """
+        Prepare package source files
```
mopsfelder: Perhaps we can briefly explain what actions are done in order to prepare the sources.  Example:

```python
def prepare_sources(self, package):
    """
    Create directory structure, create a tarball with the source code,
    and copy files to chroot.
    """
```

lib/mockbuilder.py
```diff
@@ -174,12 +219,24 @@ def _install_external_dependencies(self, package):
...
             os.path.abspath(CONF.get('common').get('work_dir')), 'mock_build',
             self.timestamp, package.name)
         os.makedirs(self.build_dir)
 
     def _destroy_build_directory(self):
+        """
+        Destroy build directory
+
+        Args:
+            package (str): package name
```
mopsfelder: There is no `package` parameter to be documented.

lib/mockbuilder.py
```diff
@@ -69,6 +69,31 @@ def build(self, package):
         LOG.info("%s: Starting build process" % package.name)
         self._build_srpm(package)
         self._install_external_dependencies(package)
+        self._build_rpm(package)
+
+        msg = "%s: Success! RPMs built!" % (package.name)
+        self._copy_rpms(self.build_dir, package.build_cache_dir)
+        LOG.info(msg)
```
ghost: Could you move this log to inside `_build_rpm`? It does not make much sense to surround `_copy_rpms` with these lines.

lib/mockbuilder.py
```diff
@@ -65,6 +71,12 @@ def initialize(self):
         utils.run_command(cmd)
 
     def build(self, package):
+        """
+        Build RPM package and its subpackages
+
+        Args:
+            package (str): package name
```
ghost: This is not a `str`, it's an `RPM_Package` object.
This applies for all methods of this class.

ghost (changes requested):

farosas:
LGTM

start build

lib/mockbuilder.py
alanoe: Done

lib/mockbuilder.py
alanoe: Done

lib/mockbuilder.py
alanoe: Done

lib/mockbuilder.py
alanoe: Done

ghost (approved):

PR #237, opened 2017-04-03
Use %{dist} macro in metapackage spec file requirements

ghost:


lib/metapackage.py
```diff
@@ -5,9 +5,13 @@
 from lib import exception
 from lib import packages_manager
 from lib import rpm_package
+from lib import utils
 from lib.versions_repository import replace_file_section
 
 LOG = logging.getLogger(__name__)
+DIST_MACRO_NAME = "%{dist}"
+DIST_MACRO_VALUE = utils.run_command(
+    "rpm --eval {}".format(DIST_MACRO_NAME)).strip()
```
mopsfelder, line 14: This will evaluate the `%dist` macro on the host system, not inside mock environment, where packages are actually built.

Don't you think we should eval `%dist` inside mock environment?

ghost: No. The `rpmspec` spec command that is executed when `SpecFile.query_tag` is called is executed in the host, so the returned release that is being replaced with this patch matches the host system one. Also, there is no chroot when `update-versions` or `update-metapackage` are executed.

You made me realize that, previously, if the host had a different dist tag than the chroot, the metapackages would require packages with a different release than the ones built. This patch is fixing that.

farosas (approved):

PR #241, opened 2017-04-11
Add support for pungi arguments

mopsfelder:
This series of patches add support to specify pungi arguments during the build of the iso image.  It also handles invalid git repositories properly, in case a non-git repository exists at the repository path.

### DIFF
--- a/lib/config.py
+++ b/lib/config.py
@@ -72,6 +72,10 @@ MOCK_ARGS = {
     ('--mock-args',):
         dict(help='Arguments passed to mock command', default=''),
 }
+PUNGI_ARGS = {
+    ('--pungi-binary',): dict(help='Pungi binary path'),
+    ('--pungi-args',): dict(help='Arguments passed to pungi command'),
+}
 RELEASE_NOTES_ARGS = {
     ('--release-notes-repo-url',):
         dict(help='Release notes repository URL'),
@@ -149,7 +153,7 @@ SUBCOMMANDS = [
     ('update-versions-readme', 'Update the supported software versions table',
         [PUSH_REPO_ARGS, DISTRO_ARGS, BUILD_REPO_ARGS]),
     ('build-iso', 'Build ISO image',
-        [ISO_ARGS, MOCK_ARGS, BUILD_ARGS]),
+        [ISO_ARGS, MOCK_ARGS, PUNGI_ARGS, BUILD_ARGS]),
 ]
 
 
--- a/lib/iso_builder.py
+++ b/lib/iso_builder.py
@@ -42,6 +42,8 @@ class MockPungiIsoBuilder(object):
         (_, _, self.arch) = distro_utils.detect_distribution()
         self.mock_binary = self.common_config.get('mock_binary')
         self.mock_args = self.config.get('mock_args') or ""
+        self.pungi_binary = self.common_config.get('pungi_binary') or "pungi"
+        self.pungi_args = self.config.get('pungi_args') or ""
 
     def _run_mock_command(self, cmd):
         distro = distro_utils.get_distro(
@@ -141,8 +143,9 @@ class MockPungiIsoBuilder(object):
 
     def _build(self):
         LOG.info("Building ISO")
-        build_cmd = ("pungi -c %s --nosource --nodebuginfo --name %s --ver %s" %
-                    (self.config.get('automated_install_file'),
+        build_cmd = ("%s %s -c %s --name %s --ver %s" %
+                    (self.pungi_binary, self.pungi_args,
+                     self.config.get('automated_install_file'),
                      self.distro, self.version))
         self._run_mock_command("--shell '%s'" % build_cmd)
 
--- a/lib/repository.py
+++ b/lib/repository.py
@@ -56,9 +56,27 @@ def get_git_repository(remote_repo_url, parent_dir_path, name=None):
         name = os.path.basename(os.path.splitext(url_parts.path)[0])
 
     repo_path = os.path.join(parent_dir_path, name)
-    if os.path.exists(repo_path):
-        MAIN_REMOTE_NAME = "origin"
+    try:
         repo = GitRepository(repo_path)
+    except git.exc.InvalidGitRepositoryError:
+        # This exception will be thrown if the Git repository has an
+        # invalid format, which could be due to the repository
+        # changing from SVN to Git, for example.  A way to proceed is
+        # removing the directory and cloning from scratch.
+        message = ("The Git repository at '{repo_path}' has an invalid format. "
+                   "If you are unsure of what to do, remove it and try again."
+                   .format(repo_path=repo_path))
+        LOG.error(message)
+        raise
+    except git.exc.NoSuchPathError:
+        message = ("Repository path '{repo_path}' does not exist."
+                   .format(repo_path=repo_path))
+        LOG.debug(message)
+        CONF = config.get_config().CONF
+        proxy = CONF.get('common').get('http_proxy')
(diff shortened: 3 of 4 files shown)


lib/repository.py
```diff
@@ -48,7 +49,7 @@ def get_git_repository(remote_repo_url, parent_dir_path):
     name = os.path.basename(os.path.splitext(url_parts.path)[0])
 
     repo_path = os.path.join(parent_dir_path, name)
-    if os.path.exists(repo_path):
+    try:
```
ghost: Please leave only the GitRepository creation line in the try block to make it clear where the exception can be raised. This makes it easier to understand the code. The code that follows it could be moved either to the `else` clause or to outside the `try`statement. See my suggestion in the last comment.

lib/repository.py
```diff
@@ -17,6 +17,7 @@
 import os
 import urlparse
 import utils
+import shutil
```
ghost: This commit is not related to this PR. I'd add it as the first commit of https://github.com/open-power-host-os/versions/pull/171, as its necessary to pass that build.

lib/repository.py
```diff
@@ -63,7 +64,14 @@ def get_git_repository(remote_repo_url, parent_dir_path):
                       .format(name=name, url=remote_repo_url))
             repo.create_remote(MAIN_REMOTE_NAME, remote_repo_url)
         return repo
-    else:
+    except git.exc.InvalidGitRepositoryError:
+        # This exception will be thrown if the git repository has an
+        # invalid format, which can be due a previous svn repository
+        # at the repo_path directory.  In this case, we remove the old
+        # gcc repository and clone from scratch.
```
ghost: This could have other causes, I'm not sure mentioning gcc is necessary here. It could be mentioned in the commit message though. I suggest:
```This exception will be thrown if the Git repository has an invalid format, which could be due to the repository changing from SVN to Git, for example. The safer way to proceed is to remove the directory and re-clone from scratch.```

File: lib/repository.py [hunk 4]
```diff
@@ -63,7 +64,14 @@ def get_git_repository(remote_repo_url, parent_dir_path):
...
+        LOG.info("Repository path '%{repo_path}' has invalid format, removing "
+                 "and cloning from scratch".format(repo_path=repo_path))
+        shutil.rmtree(repo_path)
```
ghost: A `git.exc.NoSuchPathError` could also be raised if the path does not exist. We would have to execute the same clone action in this case.

To avoid code duplication, I suggest the following structure:
```python
repo = None
try:
  repo = GitRepository(repo_path)
except git.exc.InvalidGitRepositoryError:
  remove path
except git.exc.NoSuchPathError:
  pass

if repo is None:
  return GitRepository.clone_from
else:
  create main remote
  return repo
```

lib/iso_builder.py
```diff
@@ -42,6 +42,8 @@ def __init__(self, config):
         (_, _, self.arch) = distro_utils.detect_distribution()
         self.mock_binary = self.common_config.get('mock_binary')
         self.mock_args = self.config.get('mock_args') or ""
+        self.pungi_binary = self.common_config.get('pungi_binary') or "pungi"
```
ghost, line 45: I don't think this "default" should be here. I think we should fail if the value is not present in the config file or command line. You're just mimicking the behavior above though, so let's leave this discussion for another time. Alan's PR might fix this.

lib/config.py
```diff
@@ -72,6 +72,12 @@
     ('--mock-args',):
         dict(help='Arguments passed to mock command', default=''),
 }
+PUNGI_ARGS = {
+    ('--pungi-binary',):
+        dict(help='Pungi binary path', default='/usr/bin/pungi'),
+    ('--pungi-args',):
+        dict(help='Arguments passed to pungi command', default=''),
```
ghost: Those defaults will never be used, since the `config.yaml` file takes precedence. You may remove them. This is done for other options in Alan's PR.

ghost (changes requested):
I'd squash all three commits pertaining to pungi args into one, but since they do not break anything if applied alone, you may do as you wish.

lib/repository.py
```diff
@@ -48,9 +48,27 @@ def get_git_repository(remote_repo_url, parent_dir_path):
...
         repo = GitRepository(repo_path)
+    except git.exc.InvalidGitRepositoryError:
+        # This exception will be thrown if the Git repository has an
+        # invalid format, which could be due to the repository
+        # changing from SVN to Git, for example.  A way to proceed is
+        # removing the directory and cloning from scratch.
+        message = ("Repository path '{repo_path}' has invalid format. "
+                   "If you are unsure, remove it and try again."
+                   .format(repo_path=repo_path))
+        LOG.exception(message)
```
ghost: `LOG.exception` will print the stack trace. Since you're re-raising the exception, it should be printed anyway outside of here (possibly with the program abortion), so it's better to use `LOG.error` to avoid duplicating it.

lib/repository.py
```diff
@@ -48,9 +48,27 @@ def get_git_repository(remote_repo_url, parent_dir_path):
...
+        raise
+    except git.exc.NoSuchPathError:
+        message = ("Repository path '{repo_path}' does not exist."
+                   .format(repo_path=repo_path))
+        LOG.exception(message)
```
ghost: I don't think it's a good idea to log with `ERROR` level within an expected code flow. I think `LOG.debug` would be enough, since the `clone_from` method already logs something to the user.

lib/repository.py
```diff
@@ -48,9 +48,27 @@ def get_git_repository(remote_repo_url, parent_dir_path):
...
-        MAIN_REMOTE_NAME = "origin"
+    try:
         repo = GitRepository(repo_path)
+    except git.exc.InvalidGitRepositoryError:
+        # This exception will be thrown if the Git repository has an
+        # invalid format, which could be due to the repository
+        # changing from SVN to Git, for example.  A way to proceed is
+        # removing the directory and cloning from scratch.
+        message = ("Repository path '{repo_path}' has invalid format. "
+                   "If you are unsure, remove it and try again."
```
ghost: I'd change this slightly to clarify that the repository has an invalid format, not the path:
```
message = ("The Git repository in path '{repo_path}' has an invalid format. "
           "If you are unsure of what to do, remove this path and try again."
```

lib/repository.py
mopsfelder: I've listed this pull as a dependency of https://github.com/open-power-host-os/versions/pull/171.  Good point!

lib/repository.py
mopsfelder: Done.

lib/repository.py
mopsfelder: Done.

lib/repository.py
mopsfelder: Done.

lib/iso_builder.py
mopsfelder: I removed the `default=` from options.

lib/config.py
mopsfelder: I removed the `default=` from options.

lib/repository.py
mopsfelder: Done.

lib/repository.py
mopsfelder: Done.

lib/repository.py
mopsfelder: Done.

farosas (approved):

ghost (commented):
Just rebase on top of master for the test to pass.

ghost (approved):

PR #242, opened 2017-04-12
Fix versions repository paths

ghost:
It was being obtained from the basename of the URL in some places, which
is no longer true.

Create function to get versions repository.

Regression introduced in commit add1e7be82c7724ad6d21c8b6af46d09a7d10863

### DIFF
--- a/lib/packages_manager.py
+++ b/lib/packages_manager.py
@@ -18,8 +18,8 @@ import os
 
 from lib import config
 from lib import exception
-from lib.constants import REPOSITORIES_DIR
 from lib.package import Package
+from lib.versions_repository import get_versions_repository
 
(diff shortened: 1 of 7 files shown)


lib/packages_manager.py
```diff
@@ -50,7 +47,7 @@ def prepare_packages(self, packages_class=Package,
             self.packages.append(package)
 
 
-def discover_packages():
+def discover_packages(versions_repo_path):
```
farosas: Isn't it simpler if this modules just calls `versions_repository.get_versions_repo`?

ghost: Done.

farosas (approved):

PR #245, opened 2017-04-17
Make versions repository refspecs configurable

ghost:
By making refspecs configurable and fetching the GitHub pull requests by default, we are able to simplify our infrastructure scripts.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -7,6 +7,7 @@ common:
  packages_metadata_repo_branch: 'master'
  packages_metadata_repo_refspecs:
    - '+refs/heads/*:refs/remotes/origin/*'
+   - '+refs/pull/*:refs/remotes/origin/pr/*'
  http_proxy: ''
  distro_name: 'CentOS'
  distro_version: '7'
--- a/config.yaml
+++ b/config.yaml
@@ -5,6 +5,8 @@ common:
  log_size: 2097152
  packages_metadata_repo_url: "https://github.com/open-power-host-os/versions.git"
  packages_metadata_repo_branch: 'master'
+ packages_metadata_repo_refspecs:
+   - '+refs/heads/*:refs/remotes/origin/*'
  http_proxy: ''
  distro_name: 'CentOS'
  distro_version: '7'
--- a/lib/repository.py
+++ b/lib/repository.py
@@ -26,6 +26,7 @@ from lib import exception
 
 CONF = config.get_config().CONF
 LOG = logging.getLogger(__name__)
+MAIN_REMOTE_NAME = "origin"
 
 
 class PushError(Exception):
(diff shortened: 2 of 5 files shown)


config.yaml
```diff
@@ -5,6 +5,8 @@ common:
  log_size: 2097152
  packages_metadata_repo_url: "https://github.com/open-power-host-os/versions.git"
  packages_metadata_repo_branch: 'master'
+ packages_metadata_repo_ref_specs:
```
farosas: I'd use `refspec` in one word since that is what git uses. It's also easier on people not acquainted with the term. Two words raises the question if/which of them are abbreviations (reference specification? referee's spectacles?)

lib/repository.py
```diff
@@ -129,15 +129,23 @@ def __init__(self, repo_path, *args, **kwargs):
...
+    def checkout(self, ref_name, ref_specs=None):
         """
         Check out the reference name, resetting the index state.
         The reference may be a branch, tag or commit.
+
+        Args:
+            ref_name (str): name of the reference. May be a branch, tag,
+                commit ID, etc.
+            ref_specs ([str]): pattern mappings from remote references to
+                local references, as specified by Git.
```
farosas: maybe it's worth it to include https://git-scm.com/book/id/v2/Git-Internals-The-Refspec ?

lib/repository.py
```diff
@@ -134,17 +134,10 @@ def checkout(self, ref_name):
...
-            except git.exc.GitCommandError:
-                LOG.debug("Failed to fetch %s remote for %s"
-                          % (remote.name, self.name))
-                pass
-            else:
-                LOG.info("Fetched changes for %s" % remote.name)
+        LOG.info("%(name)s: Fetching repository remote %(remote)s"
+                 % dict(name=self.name, remote=MAIN_REMOTE_NAME))
+        main_remote = self.remote(MAIN_REMOTE_NAME)
+        main_remote.fetch()
```
farosas: I think the `try: except` could stay, though

ghost: Do you mean leaving the debug message and re-raising the exception or allowing the command to fail?
We allowed the command to fail previously because we didn't know which was the remote that should be fetched and some of them might be unaccessible or invalid. Now that we only the "origin" remote, whose URL and refspecs are configurable by the user, I don't see a reason to let it fail. The default options should *never* fail, even if the URL is changed and PR refspecs are not found.

config.yaml
olavphibm: Done,

ghost: Done. Leaving the spectacles out of the question.

lib/repository.py
farosas: The former.

ghost: Done.

farosas (approved):

PR #246, opened 2017-04-19
Log empty packages

YasminBeatriz:
Print an error when there are no packages to build.

### DIFF
--- a/lib/exception.py
+++ b/lib/exception.py
@@ -70,6 +70,11 @@ class PackageDescriptorError(PackageError):
     error_code = 18
 
 
+class PackageDirectoryIsEmpty(PackageError):
+    DEFAULT_MESSAGE = "No packages to build. Packages directory is empty."
+    error_code =  19
+
+
 class RepositoryError(BaseException):
     DEFAULT_MESSAGE = (
         "Failed to setup %(repo_name)s's repository at %(repo_path)s.")
--- a/lib/packages_manager.py
+++ b/lib/packages_manager.py
@@ -85,4 +85,7 @@ def discover_packages():
         LOG.error("No packages found in versions repository directory")
         raise
 
+    if not package_list:
+    	raise exception.PackageDirectoryIsEmpty()
+
     return package_list


lib/exception.py
```diff
@@ -70,6 +70,16 @@ class PackageDescriptorError(PackageError):
...
 
 
+class PackageDirectoryIsEmpty(PackageError):
+    DEFAULT_MESSAGE = "No packages to build. Packages directory is empty."
+    error_code = 19
+
+
+class PackageDirectoryIsEmpty(PackageError):
+    DEFAULT_MESSAGE = "No packages to build. Packages directory is empty."
+    error_code = 19
```
mopsfelder: The class PackageDirectoryIsEmpty is duplicated.

lib/packages_manager.py
```diff
@@ -84,5 +84,8 @@ def discover_packages():
     except OSError:
         LOG.error("No packages found in versions repository directory")
         raise
+        
+    if not package_list:
+    	raise exception.PackageDirectoryIsEmpty()
```
mopsfelder: I think it should also inform user what to do via `LOG.error` and then raise the exception.

mopsfelder (changes requested):

lib/exception.py
YasminBeatriz: This might have happened when I was fixing the merge conflicts. Thank you for noticing.

lib/packages_manager.py
```diff
@@ -85,4 +85,8 @@ def discover_packages():
         LOG.error("No packages found in versions repository directory")
         raise
 
+    if not package_list:
+	LOG.error("No packages to build. Packages directory is empty")
```
ghost: @mopsfelder We already log the stack trace for `BaseException`s in `host_os.py`, which includes the exception custom message. This will duplicate this line in the log and stdout, so it's unnecessary.

ghost:
start tests

ghost (approved):

PR #249, opened 2017-04-27
Log base directory not found

YasminBeatriz:
Created `FilesToValidateNotFound()` exception class in `exception.py`

Throw new exception from `utils.py` script, at `recursive_glob()` function, every time `validate_*.py` scripts are called and versions repository containing the .spec/.yaml files is not found.

### DIFF
--- a/lib/exception.py
+++ b/lib/exception.py
@@ -95,4 +95,12 @@ class TimeoutError(BaseException):
         "Timeout failure on %(func_name)s after %(num_attempts)s attempts. "
         "Initial timeout: %(initial_timeout)s, final timeout: "
         "%(final_timeout)s.")
+    # Subclass errors are in the form 0b0101xxx
     error_code = 40
+
+class FilesToValidateNotFound(BaseException):
+    DEFAULT_MESSAGE = (
+        "Files to be validated not found in base directory.")
+    # Subclass errors are in the form 0b0110xxx
+    error_code = 48
+
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -20,11 +20,15 @@ import os
 import subprocess
 import time
 
-
 from lib import exception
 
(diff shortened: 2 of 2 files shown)


lib/utils.py
```diff
@@ -177,11 +175,16 @@ def recursive_glob(directory, pattern):
...
+
+    for root, _, filenames in os.walk(directory, onerror=base_directory_not_found_error):
         for filename in fnmatch.filter(filenames, pattern):
             matches.append(os.path.join(root, filename))
+
     return matches
 
+def base_directory_not_found_error(err):
+    if err.errno == errno.ENOENT:
+        raise exception.FilesToValidateNotFound()
```
olavph: You also have to re-raise the exception if it's another kind of error.

olavph: I'd define this function before the `retry_on_error` function. Albeit not an error in Python, it is a little bit weird to use a function that is defined afterwards.

lib/exception.py
```diff
@@ -81,6 +81,10 @@ class RepositoryError(BaseException):
     # Subclass errors are in the form 0b0011xxx
     error_code = 24
 
+class FilesToValidateNotFound(BaseException):
+    DEFAULT_MESSAGE = (
+        "Files to be validated not found in base directory.")
+    error_code = 25
```
olavph: Please use `error_code = 48` (0b00110000) and place this at the end of file, to keep consistency with other error binary codes. `25` is matching the prefix of a `RepositoryError`

olavph (changes requested):

lib/utils.py
YasminBeatriz: To re-raise the exception, shoud I just put `raise err` at the end of the function then?

olavph (approved):

PR #251, opened 2017-04-28
Define configuration in a single place

olavph:
This facilitates the addition and edition of options, which can all be set either through a configuration file or via command line parameters.

### DIFF
--- a/config/metadata.yaml
+++ b/config/metadata.yaml
@@ -1,0 +1,267 @@
+options:
+  architecture:
+    help: Target architecture
+    default: ppc64le
+  automated_install_file:
+    help: Path of a kickstart file, used to automate the installation of an
+      RPM-based Linux distribution
+    default: host-os.ks
+  commit_updates:
+    help: Commit file updates to local repository
+    default: True
+    action: store_false
+  distro_name:
+    help: Base Linux distribution name
+    default: CentOS
+  distro_repos_urls:
+    help: Base Linux distribution yum repositories URLs
+    default:
+      base: http://mirror.centos.org/altarch/7/os/ppc64le/
+      updates: http://mirror.centos.org/altarch/7/updates/ppc64le/
+      epel: http://download.fedoraproject.org/pub/epel/7/ppc64le/
+      extras: http://mirror.centos.org/altarch/7/extras/ppc64le/
+  distro_version:
+    help: Base Linux distribution version
+    default: '7'
+  force_rebuild:
+    help: Force the rebuild of packages. The default is to only build packages
+      when they have updated files since the last build results.
+    default: False
+    action: store_true
+  http_proxy:
+    help: HTTP proxy URL
+    default: ''
+  installable_environments:
+    help: An environment is a group of package groups. Those
+      installable environments will be available at the
+      "Sofware selection" screen in the ISO installer, along with the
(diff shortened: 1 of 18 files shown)


olavph:
build iso

olavph:
start iso

tests/unit/config_tests.py
```diff
@@ -46,13 +46,19 @@ def test_parse_arguments_list_WithLongArgument_ShouldParseArgumentValue(self, ar
         (['build-packages'], 'packages_metadata_repo_url',
          'https://github.com/open-power-host-os/versions.git'),
         (['build-packages'], 'packages_metadata_repo_branch', 'master'),
-        (['build-packages'], 'mock_args', ''),
+        (['build-packages'], 'mock_args', '--enable-plugin=tmpfs '
+         '--plugin-option=tmpfs:keep_mounted=True --plugin-option='
+         'tmpfs:max_fs_size=32g --plugin-option=tmpfs:required_ram_mb=39800 '
+         '--verbose'),
```
farosas: Aren't these changes out of place?

config/metadata.yaml
```diff
@@ -0,0 +1,339 @@
...
+    option_strings:
+    - --verbose
+    - -v
+    help: Set the scripts to be verbose
+    default: False
+    action: store_true
+  work_dir:
+    option_strings:
+    - --work-dir
+    - -w
```
farosas: Most of the options don't have short option strings, so what about making the default behavior to take the key name and add an optional `short_option_string(s)` key?:

```
work_dir:
    short_option_string: '-w' 
    help: ...
```

*I'm aware this patch-set is turning into a rabbit hole, so feel free to ignore the suggestion*

config/metadata.yaml
```diff
@@ -0,0 +1,339 @@
...
+    options:
+    - commit_updates
+    - packages_metadata_repo_branch
+    - packages_metadata_repo_refspecs
+    - packages_metadata_repo_url
+    - push_repo_branch
+    - push_repo_url
+    - push_updates
+    - updater_email
+    - updater_name
```
farosas, line 267: Minimum repetition, nice. =)

lib/config.py
```diff
@@ -163,6 +37,10 @@ def get_config():
         config_parser.parse()
     return config_parser
 
+# disable argument abbreviation in Python's ArgumentParser class
+class ArgumentParserWithoutAbbrev(argparse.ArgumentParser):
+    def _get_option_tuples(self, option_string):
+        return []
```
farosas: Why do we want this?

farosas (approved):

tests/unit/config_tests.py
olavph: The defaults in `config.py` were not matching the values in `host_os.yaml` for `mock_args`, so the tests would fail if not updated.

lib/config.py
olavph: I believed it was to avoid relying on an option short name (e.g. --mock-a[rgs]) that could be later cause ambiguity with another (e.g. --mock-add), but argparse already takes care of that by raising an exception. Removing.

config/metadata.yaml
olavph: Done.

PR #252, opened 2017-05-03
Log progress of validation scripts

YasminBeatriz:
Calling `LogHelper()` method in both `validate_*.py` scripts so I can use `LOG` variable like the other python files in this repo do.

### DIFF
--- a/scripts/validate_rpm_specs.py
+++ b/scripts/validate_rpm_specs.py
@@ -17,12 +17,14 @@
 import argparse
 import os
 import sys
+import logging
 
 repo_root_dir = os.path.realpath(os.path.join(
     os.path.dirname(__file__), os.pardir))
 sys.path.insert(0, repo_root_dir)
 
 from lib import exception
+from lib import log_helper
 from lib.utils import is_package_installed
 from lib.utils import recursive_glob
 from lib.utils import run_command
@@ -39,12 +41,12 @@ def validate_rpm_spec(spec_file_path):
     """
 
     try:
+        LOG.info("Verifying file %s" % spec_file_path)
         run_command("rpmlint -f .rpmlint -v %s" % spec_file_path)
     except exception.SubprocessError as e:
         #pylint: disable=no-member
-        print("validation of RPM specification file %s failed, output: %s" % (spec_file_path, e.stdout))
+        LOG.exception("validation of RPM specification file %s failed, output: %s" % (spec_file_path, e.stdout))
         return False
-
     return True
 
 
@@ -65,6 +67,11 @@ def validate_rpm_specs(base_dir):
         if not validate_rpm_spec(_file):
             valid = False
 
+    if valid:
+        LOG.info("Validation completed successfully.")
+    else:
+        LOG.info("Validation failed.")
+
     return valid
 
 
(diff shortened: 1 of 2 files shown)


scripts/validate_rpm_specs.py
```diff
@@ -17,16 +17,22 @@
...
 import os
 import sys
+import logging
 
 repo_root_dir = os.path.realpath(os.path.join(
     os.path.dirname(__file__), os.pardir))
 sys.path.insert(0, repo_root_dir)
 
 from lib import exception
+from lib import config
```
olavph: This import is not being used. Remove it.

scripts/validate_rpm_specs.py
```diff
@@ -39,12 +45,12 @@ def validate_rpm_spec(spec_file_path):
     """
 
     try:
+        LOG.info("Verifying file %s" %spec_file_path)
```
olavph: Add a single whitespace after the `%` operator.

scripts/validate_rpm_specs.py
```diff
@@ -39,12 +45,12 @@ def validate_rpm_spec(spec_file_path):
...
         run_command("rpmlint -f .rpmlint -v %s" % spec_file_path)
     except exception.SubprocessError as e:
         #pylint: disable=no-member
         print("validation of RPM specification file %s failed, output: %s" % (spec_file_path, e.stdout))
```
olavph: Now that we have proper logging, you can replace this print with `LOG.exception`. This will print the exception stack trace.
The same should be done for all prints, using `LOG.error` if there's no exception involved.

scripts/validate_rpm_specs.py
```diff
@@ -89,4 +100,5 @@ def parse_cli_options():
         sys.exit(1)
     args = parse_cli_options()
     if not validate_rpm_specs(args.rpm_specs_base_dir):
+        LOG.info("Packages not valid.")
```
olavph: I'd log a more precise message, like "RPM spec files contain errors."
A similar message could be logged in the YAML validation error.

scripts/validate_yamls.py
```diff
@@ -39,6 +44,7 @@ def validate_yaml(yaml_file_path):
     """
 
     try:
+        LOG.info("Verifying file %s" %yaml_file_path)
```
olavph: Add a whitespace after the `%` operator.

scripts/validate_rpm_specs.py
```diff
@@ -17,16 +17,22 @@
...
 sys.path.insert(0, repo_root_dir)
 
 from lib import exception
+from lib import config
+from lib import log_helper
 from lib.utils import is_package_installed
 from lib.utils import recursive_glob
 from lib.utils import run_command
 
+log_helper.LogHelper()
```
olavph: I think it's better to place this in the `__main__` section, like it's done in `host_os.py` to avoid having code at the module context level.

olavph (changes requested):

YasminBeatriz:
start build

olavph (approved):

farosas (approved):

PR #253, opened 2017-05-03
Update ISO checksum filename

mopsfelder:
Depends on https://github.com/open-power-host-os/infrastructure/pull/70

Currently, the ISO filename and its checksum file have different formats. An
example follows:

OpenPOWER-Host_OS-170503-ppc64le-CHECKSUM
OpenPOWER-Host_OS-DVD-ppc64le-170503.iso

Appending ".sha256" to the ISO filename to constitute the checksum file makes
filenames more consistent. An example follows:

OpenPOWER-Host_OS-DVD-ppc64le-170503.iso
OpenPOWER-Host_OS-DVD-ppc64le-170503.iso.sha256

### DIFF
--- a/lib/iso_builder.py
+++ b/lib/iso_builder.py
@@ -156,8 +156,7 @@ class MockPungiIsoBuilder(object):
         utils.force_symlink(self.timestamp, latest_dir)
 
         iso_file = "%s-DVD-%s-%s.iso" % (self.distro, self.arch, self.version)
-        checksum_file = ("%s-%s-%s-CHECKSUM" %
-                         (self.distro, self.version, self.arch))
+        checksum_file = "%s.sha256" % iso_file
         iso_dir = "/%s/%s/iso" % (self.version, self.arch)
         iso_path = os.path.join(iso_dir, iso_file)
         checksum_path = os.path.join(iso_dir, checksum_file)


lib/iso_builder.py
```diff
@@ -156,8 +156,7 @@ def _save(self):
         utils.force_symlink(self.timestamp, latest_dir)
 
         iso_file = "%s-DVD-%s-%s.iso" % (self.distro, self.arch, self.version)
-        checksum_file = ("%s-%s-%s-CHECKSUM" %
-                         (self.distro, self.version, self.arch))
+        checksum_file = "%s-CHECKSUM" % iso_file
```
olavph: That's better, but this uppercase suffix still looks weird to me. What about appending `.md5` or whatever the hash algorithm is?

olavph (approved):

lib/iso_builder.py
mopsfelder: It works for me.  I'm adding this change.

farosas (approved):

farosas:
We need a change in the infrastructure repo before merging this:
https://github.com/open-power-host-os/infrastructure/blob/master/jenkins_jobs/upload_iso/script.sh#L17

mopsfelder:
Nice catch, @fabianorosas. I'll update it accordingly.

olavph (approved):

PR #255, opened 2017-05-08
Fix typo in scheduler.py

YasminBeatriz:
Replaced _"dependecies"_ by _"dependencies"_  in `scheduler.py`. 
Replaced _"withouth"_ by _"without"_ in `/builds/workspace/repositories/gcc/gcc/sched-deps` (untracked).

Some other untracked files were also affected by the replace commands.

### DIFF
--- a/lib/scheduler.py
+++ b/lib/scheduler.py
@@ -71,7 +71,7 @@ class Scheduler(object):
         """
 
         self.packages = packages
-        LOG.info("Scheduling packages and their dependecies: %s" % packages)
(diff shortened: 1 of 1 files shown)


olavph (approved):
There's no need to mention untracked files in the PR.
Also, `/builds/workspace/repositories/gcc/gcc/sched-deps` belongs to the gcc repository, and will probably be overwritten in your next build.

YasminBeatriz:
Yes that's what I thought. I just wanted to mention this other word that was mistyped too.

PR #256, opened 2017-05-08
Documentation refactoring

farosas:


farosas:
Hyperlinks might not work in some browsers. I *think* it's an issue with GitHub; verified this with other repositories too.

README.md
```diff
@@ -14,64 +14,51 @@ For more information about Host OS check: https://open-power-host-os.github.io/
...
-```
-
-* Build all software
-
-```
-./host_os.py --verbose build-packages
-```
+  ```
+  # to build all packages
+  ./host_os.py --verbose build-packages
```
olavph: I think it's better to break this in two blocks (with ```) so it doesn't have a feeling of continuation.

You're also leaving the `--verbose` option in the first command, but not in the second. I think it could be removed.

README.md
```diff
@@ -7,12 +7,15 @@ script that provides software well packaged and designed for the
 
 For more information about Host OS check: https://open-power-host-os.github.io/
 
-## Supported GNU/Linux distributions
+* [Quick Start](#quick-start)
+* [Building](docs/build.md#building)
+  * [iso](docs/building_an_iso.md)
+  * [custom software](docs/build.md#build-custom-software)
+  * [previous Host OS versions](docs/build_old_versions.md)
```
olavph: Uppercase? At least for ISO.

docs/build.md
```diff
@@ -0,0 +1,8 @@
+* Build custom software
+
+The build code keeps a cache of all packages source code (Git/Mercurial/SVN)
+repositories so that they do not need to be downloaded on every build. If it is
+desired to change the source code of a package, you may edit the package local
+repository content. To build the package from this modified local repository,
+either use a file:// URL pointing to it in the package metadata (YAML)
+file or disable package repositories update in config/host_os.yaml.
```
olavph: Unrelated to this patch. I don't think this should be the recommended way of building custom software. We should improve this file, giving more precise instructions on how to build from a custom local or remote repository.

docs/build.md
```diff
@@ -1,8 +1,126 @@
...
+information about the components to build, such as URLs, versions,
+commit IDs, dependencies and distro-specific build parameters.
+
+During the build process, the build script checks out the metadata
+(versions) repository, obtains the information necessary to build the
+package and starts an instance of
+[mock](https://github.com/rpm-software-management/mock/wiki) that
+builds the package in a chroot, taking care of its build dependencies.
+
+Other functionalities of the build script are:
```
olavph, line 14: Not sure if this info should be mixed with the build process. Maybe we could separate them to a "release process" doc.

docs/build.md
```diff
@@ -1,8 +1,126 @@
...
+./host_os.py build-packages --packages kernel
+```
+
+which builds a specific package or list of packages.
+
+The built packages are kept in a `result` directory (which can be
+changed with the `--result-dir` option), and are not rebuilt in
+subsequent runs of the command unless their source file have changed
+since the creation of the last build results (to avoid this, use
+--force-rebuild).
```
olavph, line 67: This is basically a copy of the quick start guide. Should we really keep this info duplicated? It will likely lead to inconsistent docs in the future.

docs/build.md
```diff
@@ -1,8 +1,126 @@
...
+## Building packages with modified metadata (source URL, commit ID, branch)
+
+The `versions` repository contains build information for each Host OS
+package such as source code URL, source code retrieval method (git,
+svn, wget), git commit ID, branch, svn revision. This information can
+be altered in a number of ways:
+
+  - by making changes to a local clone of the `versions` repository and issuing:
+    ```
+    ./host_os.py build-packages --package <pkg> --packages-metadata-repo-url file://<absolute path to versions>
```
olavph: The correct parameter is `--packages`. Isn't it better to omit this, since we are explaining the `--packages-metadata-repo-url` option? We could include the `--packages-metadata-branch` too.

docs/build.md
```diff
@@ -1,8 +1,126 @@
...
+    ```
+
+  - by making changes to `versions`, pushing them to a remote location (e.g. GitHub fork) and then:
+    ```
+    ./host_os.py build-packages --package <pkg> --packages-metadata-repo-url https://github.com/<user>/versions
+    ```
+
+  - by NOT making changes to `versions` repository, but informing them via command line to the `build-packages` command:
+    ```
+    ./host_os.py build-packages --package package_name#repo_url#reference
```
olavph: `--packages`

docs/build.md
```diff
@@ -1,8 +1,126 @@
...
+so that they do not need to be downloaded on every build. If it is
+desired to change the source code of a package, you may edit the local
+cached content.
+
+*tip: for the cached content location, check the `work_dir` variable in the `builds` repository configuration file.*
+
+To build a package from this modified local repository, either use the
+methods already [mentioned](#specifying-metadata) or disable package
+repositories update (`--no-update-packages-repos-before-build`
+option).
```
olavph: The methods above should not be used if the cached repository has been changed. The script would checkout the specified commit, possibly dropping local changes and losing commit references. I suggest:

To build a package from this modified local repository, disable package repositories update (`--no-update-packages-repos-before-build` option).

WARNING: Whenever you build without disabling package repositories update (`--no-update-packages-repos-before-build` option), changes made to the cached repositories may be lost.

docs/build.md
```diff
@@ -1,8 +1,126 @@
...
+    ./host_os.py build-packages --package package_name#repo_url#reference
+
+    # e.g.:
+    ./host_os.py build-packages --package kernel#file:///home/user/libvirt#a2436b799ff91e304aad4e590f0adb0114908780
+    ```
+
+    *tip: check `./host_os.py build-packages --help` for other ways of specifying the `--package` option*
+
+
+## Building packages with modified sources
```
olavph: The section above also builds packages with modified sources. I consider this a secondary functionality not recommended for most users.

Anyway, if we want to keep those instructions, I suggest: Building packages with locally modified sources

docs/build.md
```diff
@@ -1,8 +1,126 @@
...
+
+The build code keeps a cache of the packages' source code repositories
+so that they do not need to be downloaded on every build. If it is
+desired to change the source code of a package, you may edit the local
```
olavph: ...desired to build a package with small changes in its source code, ...

olavph (changes requested):

docs/build.md
farosas: Agreed. This feels like a hack. Is there any usefulness to "disabling package repositories update"? Otherwise, we could drop this option.

But what should be the recommended way of doing it, though? Cloning the package source code again?

docs/build.md
farosas: I get your point but I want to give an overview of what the script does so that the user doesn't get completely confused when `./host_os.py --help` shows a bunch of weird stuff. 

I don't think the other commands are useful to anyone besides us (perhaps `update-versions`?) and, as you know, IMO they shouldn't even be in the same repository. So I think having a "release process" documentation makes no sense if nobody cares about it.

But I'm open to discussion. Maybe we can find a middle ground.

*As a side note:
I don't even believe in the "release" concept anymore, maybe we could just keep it rolling, drop `update-versions` and `update-metapackage` and tweak `build-release-notes` to be a bit more process agnostic.*

docs/build.md
farosas: I'd like this file to be the authoritative source of documentation about building. The "quick start guide" will inevitably repeat some of the information here. What I could do is to make the quick start even smaller to lower the risk of inconsistencies.

docs/build.md
farosas: I'll make the suggested change to the doc, but let's discuss this further to improve the process behind it.

docs/build.md
farosas: But the `--package <pkg>#<url>#<sha1># ` option would still work, right?

docs/build.md
olavph: Ok, we can leave it here until we decide what to do with these stuff.

docs/build.md
olavph: I get it. Maybe we could remove the part that we mention `--help` and the `results` directory from the quick start guide and link this page for extended explanation and more advanced usage. The `--help` should then be mentioned here.

docs/build.md
olavph: It would work because of the way we've implemented it, but not by design. If we pass `--packages <pkg>#file://<path_to_cached_dir>#<sha1>`, the script would create a remote in the cached repository pointing to itself, which is very weird, and would indeed checkout the correct commit. In fact, any URL that we pass would work, even if the pointed repository does not contain the requested sha1.

If, for some reason, we decide to add an option to wipe the cached dir prior to checking out, it would break this and lose the user's commit.

docs/build.md
olavph: Yes, I believe the recommended way would be to have the package source code cloned. Not necessarily again, because the user might have a custom package repository even before executing their first Host OS build. The new [naive] user should not need to know where we keep the cached repositories and how we manage them.

The only usefulness of this is being able to make quick changes to sources to test something, without having to commit or clone to another location. I've never used it myself, but I see it as a nice feature for some use cases.

docs/build.md
farosas: Yes, relying on the cache modification is not a good thing. What about instructing the user to copy the cached repo to another directory or cloning it again and then using `--packages ###`? Would that be acceptable?

olavph: I prefer to instruct to clone the remote repository, as it is a more standard git practice. We could maybe instruct the user to use `--reference <path_to_cached_repo>` to speed up the clone and future fetches.

olavph (approved):

PR #257, opened 2017-05-09
Revert "Update ISO checksum filename"

mopsfelder:
Revert "Update ISO checksum filename"

This reverts commit 6c5e63750a91118db33ccff03408dcd9ba938023.

The above patch broke build-iso subcommand because the checksum file was still
being created as `*-CHECKSUM` and the wrong path was passed to mock --copyout
parameter.

Renaming the checksum file will likely happen in _save() function after files
are copied out from mock.


olavph (approved):
A small explanation on why this broke mock would be nice.

mopsfelder:
@olavph, I updated the commit message and pull request description. Let me know if more details are needed.

olavph (approved):

PR #258, opened 2017-05-09
Save netinst ISO image

mopsfelder:
The netinst image is already generated by the compose tool but it is not
currently saved as part of the resulting artifacts.

This patch saves the DVD and netinst images along with the checksum file.

### DIFF
--- a/lib/iso_builder.py
+++ b/lib/iso_builder.py
@@ -155,18 +155,12 @@ class MockPungiIsoBuilder(object):
                                   LATEST_SYMLINK_NAME)
         utils.force_symlink(self.timestamp, latest_dir)
 
-        iso_file = "%s-DVD-%s-%s.iso" % (self.distro, self.arch, self.version)
-        checksum_file = ("%s-%s-%s-CHECKSUM" %
-                         (self.distro, self.version, self.arch))
         iso_dir = "/%s/%s/iso" % (self.version, self.arch)
-        iso_path = os.path.join(iso_dir, iso_file)
-        checksum_path = os.path.join(iso_dir, checksum_file)
-        chroot_files = "%s %s" % (iso_path, checksum_path)
+        iso_files = "%s/*" % iso_dir
 
-        LOG.info("Saving ISO %s and checksum %s at %s" %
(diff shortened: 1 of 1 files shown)


lib/iso_builder.py
```diff
@@ -155,16 +155,11 @@ def _save(self):
...
 
-        iso_file = "%s-DVD-%s-%s.iso" % (self.distro, self.arch, self.version)
-        checksum_file = ("%s-%s-%s-CHECKSUM" %
-                         (self.distro, self.version, self.arch))
         iso_dir = "/%s/%s/iso" % (self.version, self.arch)
-        iso_path = os.path.join(iso_dir, iso_file)
-        checksum_path = os.path.join(iso_dir, checksum_file)
-        chroot_files = "%s %s" % (iso_path, checksum_path)
+        iso_files = "%s/*" % iso_dir
+        chroot_files = "%s" % iso_files
```
farosas: You could simply remove the `chroot_files` variable and use `iso_files` everywhere.

farosas (changes requested):

lib/iso_builder.py
mopsfelder: Done.

lib/iso_builder.py
```diff
@@ -155,18 +155,12 @@ def _save(self):
...
         utils.force_symlink(self.timestamp, latest_dir)
 
-        iso_file = "%s-DVD-%s-%s.iso" % (self.distro, self.arch, self.version)
-        checksum_file = ("%s-%s-%s-CHECKSUM" %
-                         (self.distro, self.version, self.arch))
         iso_dir = "/%s/%s/iso" % (self.version, self.arch)
-        iso_path = os.path.join(iso_dir, iso_file)
-        checksum_path = os.path.join(iso_dir, checksum_file)
-        chroot_files = "%s %s" % (iso_path, checksum_path)
+        iso_files = "%s/*" % iso_dir
```
olavph, line 159: I'd call this `iso_files_pattern` so no one confuses it with a list of files, but that's a matter of taste.

olavph (approved):

farosas (approved):

PR #262, opened 2017-05-18
Add iso version to build-iso subcommand

mopsfelder:
This adds --iso-version param to build-iso subcommand of host_os.py, allowing an
iso version to be specified to the final iso name; and shortens iso name to comply
with the 32-characters restriction in the volume id.

### DIFF
--- a/config/metadata.yaml
+++ b/config/metadata.yaml
@@ -71,6 +71,10 @@ options:
     - OpenPOWER Host OS virtualization
     - OpenPOWER Host OS virtualization management
     - OpenPOWER Host OS RAS
+  iso_version:
+    help: ISO version
+    # If empty, current date is used
+    default: ''
   keep_build_dir:
     help: Keep build directory and its logs and artifacts
     default: False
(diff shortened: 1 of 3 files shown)


lib/iso_builder.py
```diff
@@ -37,7 +37,8 @@ def __init__(self, config):
         self.result_dir = os.path.join(self.config.get('result_dir'),
             'iso', self.timestamp)
         self.distro = self.config.get("iso_name")
-        self.version = datetime.date.today().strftime("%y%m%d")
+        self.version = self.config.get("iso_version") or \
```
olavph: PEP8 recommends using parenthesis instead of backslash for wrapping lines.

farosas (approved):

config/metadata.yaml
```diff
@@ -71,6 +71,9 @@ options:
     - OpenPOWER Host OS virtualization
     - OpenPOWER Host OS virtualization management
     - OpenPOWER Host OS RAS
+  iso_version:
+    help: ISO version
```
olavph, line 75: It'd be good to mention that if left empty, the current date and time are used.

olavph (approved):

PR #264, opened 2017-05-22
Mention installing epel release

YasminBeatriz:
In `docs/install_packages.md` a little paragraph was added to remind the user to follow the steps in Quick Start, and install epel release packages before installing Host OS packages from repository.

### DIFF
--- a/docs/install_packages.md
+++ b/docs/install_packages.md
@@ -6,7 +6,8 @@ system with a
 and a yum configuration file pointing to an accessible yum repository
 containing the Host OS packages.
 
-If you've just executed the build script and want to use the local yum
+If you've just followed the [Quick Start guide](https://github.com/open-power-host-os/builds#quick-start) 
+and want to use the local yum
 repository containing the built packages, your yum configuration file
 will be available at `result/repository_config/latest`. Otherwise, you'll
 need one pointing to the remote yum repository.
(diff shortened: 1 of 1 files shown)


docs/install_packages.md
```diff
@@ -6,6 +6,9 @@ system with a
 and a yum configuration file pointing to an accessible yum repository
 containing the Host OS packages.
 
+Also, you need to follow the steps in the [Quick Start](https://github.com/open-power-host-os/builds#quick-start)
+and install epel-release packages to avoid dependency errors.
+
```
farosas: This text got a little out-of-place when considered in context with the other paragraphs, mainly regarding the statement below: "If you've just executed the build script". Consider the questions:

- Do we expect the user to follow the Quick Start or build packages in some other way? Line 9 you added implies the former, but line 12 also suggests a different approach is possible.
- If we need epel-release to avoid dependency errors (l10), how come "Yum will take care of the dependencies" (l31)?

Maybe if we just mention epel-release between lines 22-24, like so:

> Make sure EPEL is installed:
>
> ```
> sudo yum install epel-release
> ```

and then remove the fallacious line 31.

If we want to mention the Quick start guide, I think that replacing "just executed the build script" (l12) with "just followed the Quick Start guide", would be a good call.

farosas (changes requested):

farosas (approved):

olavph (approved):

PR #269, opened 2017-06-14
Do not bump metapackage release when there are no updates

YasminBeatriz:
Create `boolean` flag in `Version.update(self, user_name, user_email)` to check whether any package has actually been updated.

### DIFF
--- a/lib/subcommands/update_versions.py
+++ b/lib/subcommands/update_versions.py
@@ -82,6 +82,9 @@ class Version(object):
                 change logs
             user_email (str): email used when updating RPM specification files
                 change logs
+
+        Returns:
+            bool: Whether the package has been updated
         """
         previous_commit_id = self.pkg.sources[0]["git"]["commit_id"]
         previous_version = self.pkg.version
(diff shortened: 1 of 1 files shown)


lib/subcommands/update_versions.py
```diff
@@ -94,7 +94,7 @@ def update(self, user_name, user_email):
 
         if newest_commit_id == previous_commit_id:
             LOG.debug("%s: no changes.", self.pkg)
-            return
+            return False
```
olavph, line 99: Please mention the return in the doc string:
```
Returns:
    bool: whether the package has been updated
```

lib/subcommands/update_versions.py
```diff
@@ -227,17 +229,20 @@ def run(CONF):
...
 
     packages_names = discover_packages()
     METAPACKAGE_NAME = "open-power-host-os"
     packages_names.remove(METAPACKAGE_NAME)
 
-    update_metapackage(
-        versions_repo, distro, METAPACKAGE_NAME, packages_names,
-        updater_name, updater_email)
-    update_versions_in_readme(versions_repo, distro, packages_names)
+    if updates_available:
```
olavph: The commit should not be created if there are no updates. You could either include the rest of the function in the `if` or place this right before the call to `discover_packages()`:
```
if not updates_available:
    LOG.info("No updates in packages versions.")
    continue
```
I prefer the former to avoid indenting all those lines.

olavph (changes requested):
This is conflicting with the master branch. Please rebase on top of it to make it mergeable.

olavph (approved):

farosas (approved):

PR #273, opened 2017-06-26
Add --uniqueext to mock command in MockPungiIsoBuilder

olavph:
This is a step closer to building ISOs concurrently and avoids chroot
conflicts when builds fail.

Refactor code that deals with mock.
Fix distro object creation message being logged multiple times.

### DIFF
--- a/lib/mock.py
+++ b/lib/mock.py
@@ -1,0 +1,53 @@
+# Copyright (C) IBM Corp. 2017.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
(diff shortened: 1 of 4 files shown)


lib/mock.py
```diff
@@ -0,0 +1,49 @@
...
+from lib import config
+from lib import utils
+
+CONF = config.get_config().CONF
+
+class Mock(object):
+
+    def __init__(self, config_file, unique_extension):
+        """
+        Constructor
```
farosas: =(

farosas (approved):
Well done in creating the Mock class o/

lib/mock.py
olavph: Haha, gave it a meaningful description: `Initialize arguments common to all mock calls.`

PR #274, opened 2017-07-03
Handle missing parameters in config file

YasminBeatriz:
In case some argument is not provided by command-line, there is an extra step in `config.py` to check whether this argument exists in config file. An exception will be thrown either when the key does not exist, or when it has `None` type assigned to it. 
For optional arguments, the user should provide an empty string.

### DIFF
--- a/lib/config.py
+++ b/lib/config.py
@@ -221,7 +221,10 @@ class ConfigParser(object):
         # the subcommand
         for key, value in command_line_args.items():
             if value is None:
-                continue
+                if config.get(key) is None:
+                    raise Exception("Option {} has no value set".format(key))
+                else:
+                    continue
             if key in config:
                 target_type = type(
                     config_metadata["options"][key]["default"])


lib/config.py
```diff
@@ -221,7 +221,10 @@ def parse(self, command_line_args=None):
         # the subcommand
         for key, value in command_line_args.items():
             if value is None:
-                continue
+                if config.get(key) is None:
+                        raise Exception("Option {} has no value set".format(key))
```
olavph: We use 4 spaces for indentation, please set this in your editor.

olavph (changes requested):

lib/config.py
```diff
@@ -220,7 +220,10 @@ def parse(self, command_line_args=None):
         # Update config node with command line arguments applicable to
         # the subcommand
         for key, value in command_line_args.items():
-            if value is None:
+        if value is None:
```
olavph: This `if` is outside the `for`. This and the next 4 lines need 4 more spaces.

olavph (changes requested):

olavph (approved):

PR #282, opened 2017-07-31
Use systemd-nspawn for bind mounting

farosas:
With mock 1.4.2 using systemd-nspawn, these directories are already
mounted.

EDIT: nspawn creates them but it doesn't create any loop devices, so I'm altering the patch to bind mount /dev, but using nspawn this time.


farosas:
Remember that this must wait until the other issues with mock are fixed before being merged.

farosas:
Closing in favor of #283

PR #283, opened 2017-08-03
Nspawn fixes

farosas:


mopsfelder:
In the commit message of 59223d4897a7ebb54cd3bc4771d01bc663724b2a:

    but it is not present in our systemd version (219).

Perhaps we should be clearer that that option is not present in the systemd 219 from CentOS 7.3.

mopsfelder (approved):

olavph (approved):

PR #284, opened 2017-08-08
Add support to extraver in metapackage update code

farosas:


lib/rpm_package.py
```diff
@@ -75,7 +75,6 @@ def write_content(self):
     def query_tag(self, tag_name, extra_args=""):
         """
         Queries the spec file for a tag's value.
-        Cached content not yet written to the file is not considered.
```
olavph: This text might be somewhat misleading, but it's not wrong.
There are two caches used in this class: `_cached_tags`, which contains the tags read from the file, and `_content`, which holds the text that was yet not written to the file.
This method ignores `_content` and reads directly from the file, which could be outdated as compared to the the cached `_content`.

All that said, all methods that alter `_content` already flush it at the end, so this bypass does not currently happen. I still think it's useful to leave some kind of warning here so if an issue arises we remember this.

lib/rpm_package.py
```diff
@@ -96,9 +98,21 @@ def query_tag(self, tag_name, extra_args=""):
...
+            dummy_macros = {m : "dummy_" + m for m in unexpanded_macros}
+            extra_args += get_define_line(dummy_macros)
+
             command = "rpmspec --srpm -q --qf '%%{%s}' %s %s 2>/dev/null" % (
                     tag_name.upper(), self.path, extra_args)
             tag_value = utils.run_command(command).strip()
+
+            for macro in unexpanded_macros:
+                tag_value = tag_value.replace('dummy_%s' % macro,
+                                              '%%{%s}' % macro)
```
olavph, line 118: Just a thought: if the macro happens to be defined in the spec file, this replace will not work, since the dummy value will be overwritten by whatever is in the file. I don't think we have to worry with this use case though.

lib/metapackage.py
```diff
@@ -56,7 +56,8 @@ def replace_spec_dependencies(spec_file_path):
                 else:
                     package_evr = ""
                 release = package.spec_file.query_tag(
-                    "release", unexpanded_macros=['dist'])
+                    "release", unexpanded_macros=['dist', 'extraver'])
+                release = release.replace('extraver', '?extraver')
```
olavph: Why not place this "?" in the `query_tag` function?

farosas: Because that is a thing specific to this code. Just like you chose to not use '?' in the `dist` tag, I chose to use it for the `extraver`. The actual macro usage in the spec file may vary and `rpmspec` only see the outcome of the expansion, so there is no way (using rpmspec) to know if the original macro had the symbol or not.

olavph (approved):

PR #289, opened 2017-08-17
release-notes: create webpage post with information from info files

farosas:
Now that the build script and build pipeline output .json files with data about the build, every front-end component should get data from these files so that we have a single source of data to avoid inconsistencies.

This patch-set alters the creation of the release page posts to make use of the files instead of gathering the information itself.

### DIFF
--- a/config/metadata.yaml
+++ b/config/metadata.yaml
@@ -39,6 +39,9 @@ options:
   http_proxy:
     help: HTTP proxy URL
     default: ''
+  info_files_dir:
+    help: Path to a directory containing the build information files
+    default: result/packages/latest
   installable_environments:
     help: An environment is a group of package groups. Those
       installable environments will be available at the
@@ -209,17 +212,15 @@ commands:
     help: Build release notes
     options:
     - commit_updates
-    - packages_metadata_repo_branch
-    - packages_metadata_repo_refspecs
-    - packages_metadata_repo_url
+    - info_files_dir
     - push_repo_branch
     - push_repo_url
     - push_updates
     - release_notes_repo_branch
     - release_notes_repo_url
+    - update_packages_repos_before_build
     - updater_email
     - updater_name
-    - update_packages_repos_before_build
   build-iso:
     help: Build ISO image
     options:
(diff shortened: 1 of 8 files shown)


config/metadata.yaml
```diff
@@ -220,6 +220,7 @@ commands:
     - updater_email
     - updater_name
     - update_packages_repos_before_build
+    - info_files_path
```
olavph: Minor: This list was in alphabetical order.

config/metadata.yaml
```diff
@@ -39,6 +39,9 @@ options:
   http_proxy:
     help: HTTP proxy URL
     default: ''
+  info_files_path:
```
olavph: Minor: I'd prefer `info_files_dir`

lib/subcommands/build_release_notes.py
```diff
@@ -42,45 +36,62 @@
...
+            continue
 
-    version_milestone = read_version_and_milestone(versions_repo)
+        with open(os.path.join(info_files_path, file_name)) as json_file:
+            file_content = json.load(json_file)
+
+        node_name = file_name.replace('.json','')
+        if node_name == 'build':
+            # leave build information at the top-level
+            info.update(file_content)
```
olavph: I don't see any advantage on having a special treatment to leave this at the top level.

lib/subcommands/build_release_notes.py
```diff
@@ -42,45 +36,62 @@
...
+            header_yaml=yaml.safe_dump(release_file_info, default_flow_style=False))
         version_info_file.write(release_file_content)
 
 
-def run(CONF):
-    versions_repo = setup_versions_repository(CONF)
+def read_info_files(info_files_path):
+
+    info = {}
+    for file_name in os.listdir(info_files_path):
```
olavph: Wouldn't it be simpler to read the two files we need explicitly?
I feel like merging those dicts is just adding unnecessary complexity. We could pass both dicts to `write_version_info` (build_info and packages_info).

olavph (approved):

PR #293, opened 2017-09-15
Add parameter to set release category

olavph:
This allows us to have multiple pipelines publishing release notes.
The category is appended as the file suffix to serve as the post title
and added in the YAML metadata.

### DIFF
--- a/lib/subcommands/build_release_notes.py
+++ b/lib/subcommands/build_release_notes.py
@@ -28,7 +28,7 @@ from lib.constants import REPOSITORIES_DIR
 CONF = config.get_config().CONF
 LOG = logging.getLogger(__name__)
 
-RELEASE_FILE_NAME_TEMPLATE = "{date}-release.markdown"
+RELEASE_FILE_NAME_TEMPLATE = "{date}-{suffix}.markdown"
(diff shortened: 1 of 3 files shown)


lib/subcommands/build_release_notes.py
```diff
@@ -60,6 +62,7 @@ def write_version_info(file_path, release_date, build_info, packages_info):
     release_file_info = {
         "title": RELEASE_FILE_TITLE,
         "layout": RELEASE_FILE_LAYOUT,
+        "categories": "tags " + release_category,
```
farosas, line 65: Where does this go? I don't see it anywhere on the page.

olavph: This is used here:
https://github.com/open-power-host-os/open-power-host-os.github.io/pull/47/files#diff-8a9f2e5bb76bff2b106a74763fc92b92R2
https://github.com/open-power-host-os/open-power-host-os.github.io/pull/47/files#diff-c64156f29f3a87e1a368107231f5aa4eR6

"tags" is mentioned to keep the page listing all tags (devel+release+stable), although it is not linked anywhere.

PR #296, opened 2017-09-19
Remove sos from list of packages to update

olavph:
This package has been removed from versions repository because the
distro's version surpassed ours.


olavph:
Related to https://github.com/open-power-host-os/versions/pull/224

PR #299, opened 2017-10-09
Move VERSION file

YasminBeatriz:


olavph:
Depends on https://github.com/open-power-host-os/versions/pull/234

PR #304, opened 2017-11-03
Rename sub-command build-iso to build-images

mopsfelder:
Besides renaming build-iso to build-images, this pull adds --iso and --install-tree to build-images sub-command to build ISO and installable tree directory, respectively.

Depends on https://github.com/open-power-host-os/infrastructure/pull/123


lib/iso_builder.py
```diff
@@ -89,6 +89,7 @@ def __init__(self, config):
         (_, _, self.arch) = distro_utils.detect_distribution()
         self.pungi_binary = self.config.get('pungi_binary') or "pungi"
         self.pungi_args = self.config.get('pungi_args') or ""
+        self.build_iso = self.config.get('iso') or False
```
olavph: The `or False` is kind of redundant. It is needed in the other cases because if the config doesn't exist, `None` will be returned, and passing `None` when a string is expected leads to strange behavior (`None` is converted to `"None"`).
Passing `None` instead of a boolean is common in python, though.
If you want to leave it to be more explicit, no problem. Another option would be `bool(self.config.get('iso'))`

olavph: You could mention in the commit message that this commit is only preparing for the addition of other build-images options. As I understand it, this commit is not altering any behavior.

lib/iso_builder.py
```diff
@@ -311,5 +312,14 @@ def _save(self):
             self.mock.run_command("--copyout %s %s" %
                                   (iso_files, self.result_dir))
 
+        tree_src_dir = "/%s/%s/os" % (self.version, self.arch)
+        tree_dest_dir = self.result_dir + "/os"
```
olavph: I prefer to use `os.path.join` instead of concatenating like this.

lib/subcommands/build_images.py
```diff
@@ -21,12 +21,14 @@
 
 
 def run(CONF):
-    build_iso = CONF.get('iso')
+    build_iso = CONF.get('iso') or False
+    build_install_tree = CONF.get('install_tree') or False
```
olavph: Same as previous comment. The `or False` is optional.

lib/subcommands/build_images.py
```diff
@@ -21,12 +21,14 @@
...
 
-    if build_iso:
+    if not build_iso and not build_install_tree:
```
olavph: This could still be in the `else` block.

olavph (approved):

PR #310, opened 2018-05-15
lib/utils: check for installed packages in a more reliable way

farosas:
Some rpm packages might be installed under a different name than what
we expect. For instance:

$ rpm -qa | grep GitPython
python2-GitPython-2.1.7-2.fc26.noarch

$ rpm -q GitPython
package GitPython is not installed

This patch uses a more flexible approach to checking for installed
packages:

$ rpm -q --whatprovides GitPython
python2-GitPython-2.1.7-2.fc26.noarch

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -166,7 +166,7 @@ def is_package_installed(package_name):
         bool: if RPM package is installed
     """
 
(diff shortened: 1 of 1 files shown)


mopsfelder (approved):
LGTM

:+1:

PR #317, opened 2018-08-03
Cleanup subpackages and installable environments

mopsfelder:


lib/metapackage.py
```diff
@@ -108,32 +108,6 @@ def update_metapackage(
         metapackage_yaml_file_path, yaml_install_dependencies_string,
         YAML_START_DELIMITER)
 
-    LOG.info("Updating release package spec file")
```
farosas: There is some code in here which relates to the update of the metapackage with information from the other packages, so I'd leave this message here.

lib/metapackage.py
```diff
@@ -108,32 +108,6 @@ def update_metapackage(
...
-    with open(version_file_path) as version_file:
-        version_parts = version_file.readlines()[-1].strip().split("-")
-    new_metapackage_version = version_parts.pop(0)
-    if version_parts:
-        metapackage_milestone = version_parts.pop(0)
-    else:
-        metapackage_milestone = "%nil"
-
-    metapackage = rpm_package.RPM_Package.get_instance(
-        metapackage_name, distro)
```
farosas: We need this code.

lib/metapackage.py
mopsfelder: Done

lib/metapackage.py
```diff
@@ -109,31 +109,10 @@ def update_metapackage(
         YAML_START_DELIMITER)
```
farosas, line 108: ^ unused import up at line 5

farosas (approved):
LGTM, just please remember to make it pass validation

